PR #181, opened 2021-10-27
feat: reformat data loading

tadejsv:
Here I will develop the data-related part of reformatting, so the part that concerns the sampling and batching functions.

Stay *tuned* ;)


finetuner/tuner/dataset/__init__.py
```diff
@@ -1,8 +1,184 @@
...
+    @abc.abstractmethod
+    def __getitem__(
+        self, ind: int
+    ) -> Tuple[Union[str, np.ndarray], Union[int, Tuple[int, int]]]:
+        ...
+
+    @property
+    @abc.abstractmethod
+    def labels(self) -> Union[int, Tuple[int, int]]:
+        ...
```
bwanglzu, line 29: what is this expect label?

tadejsv: Depends on dataset - see this method for class and sessions dataset

finetuner/tuner/dataset/__init__.py
```diff
@@ -1,8 +1,184 @@
...
+        self._docs = docs
+        self._preprocess = preprocess
+
+        self._tag_labels_dict: Dict[Union[str, int], int] = {}
+        self._labels: List[int] = []
+        max_label = 0
+
+        for doc in self._docs:
+            try:
+                tag = doc.tags['finetuner']['label']
```
bwanglzu, line 63: use default tag key in helper instead of hard-coded

finetuner/tuner/dataset/__init__.py
```diff
@@ -1,8 +1,184 @@
...
+
+        return (content, label)
+
+    @property
+    def labels(self) -> List[int]:
+        """ Get the list of integer labels for all items in the dataset."""
+        return self._labels
+
+    def __len__(self) -> int:
+        return len(self._labels)
```
bwanglzu, line 97: this can be done in base class

finetuner/tuner/dataset/__init__.py
```diff
@@ -1,8 +1,184 @@
...
+
+        if self._preprocess is not None:
+            content = self._preprocess(content)
+
+        return (content, label)
+
+    @property
+    def labels(self) -> List[int]:
+        """ Get the list of integer labels for all items in the dataset."""
+        return self._labels
```
bwanglzu, line 94: this can be done in base class

finetuner/tuner/dataset/__init__.py
```diff
@@ -1,8 +1,184 @@
...
+
+        A label consists of two integers, the session ID (index of root document in
+        original document array), and the type, which is 0 if the document is the
+        anchor (root) document, 1 if it is a positive input (match), and -1 if it is
+        a negative input (match)
+        """
+        return self._labels
+
+    def __len__(self) -> int:
+        return len(self._labels)
```
bwanglzu, line 181: same as above

bwanglzu (changes requested):
not sure what is the `max_label` defined in class

finetuner/tuner/dataset/__init__.py
```diff
@@ -1,8 +1,184 @@
...
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 
 from ... import __default_tag_key__
+from ...helper import DocumentSequence
+
+
+class BaseDataset:
+    @abc.abstractmethod
```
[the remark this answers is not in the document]
tadejsv: Sure, will remove

finetuner/tuner/dataset/__init__.py
```diff
@@ -1,8 +1,184 @@
...
+    def labels(self) -> Union[int, Tuple[int, int]]:
+        ...
+
+    @abc.abstractmethod
+    def __len__(self) -> int:
+        ...
+
+
+class ClassDataset:
```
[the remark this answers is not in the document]
tadejsv: I think it would be confusing - so then all other datasets are unlabelled?

finetuner/tuner/dataset/__init__.py
```diff
@@ -1,8 +1,184 @@
...
 from ... import __default_tag_key__
+from ...helper import DocumentSequence
+
+
+class BaseDataset:
+    @abc.abstractmethod
+    def __init__(
+        self,
+        docs: DocumentSequence,
+        preprocess: Optional[Callable[[Union[str, np.ndarray]], Any]] = None,
```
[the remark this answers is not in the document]
tadejsv: No, that's not what callbacks are for - check, for example [keras callbacks](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/Callback#on_train_batch_begin), no access to data is given.

Will change the name to `preprocess_fn`

finetuner/tuner/dataset/__init__.py
tadejsv: Removed the base class as it did not really much. Besides, I want this to be different as methods in different class have different docstrings

finetuner/tuner/dataset/__init__.py
```diff
@@ -1,8 +1,158 @@
...
+
+        self._tag_labels_dict: Dict[Union[str, int], int] = {}
+        self._labels: List[int] = []
+        max_label = 0
+
+        for doc in self._docs:
+            try:
+                tag = doc.tags[__default_tag_key__]['label']
+            except KeyError as e:
+                raise KeyError(
```
bwanglzu, line 39: reminder to check Hans note on better trace in team engineering, i think it's better to use `ValueError`.

finetuner/tuner/dataset/__init__.py
```diff
@@ -1,8 +1,158 @@
...
+
+    def __getitem__(self, ind: int) -> Tuple[Union[np.ndarray, str], int]:
+        """
+        Get the (preprocessed) content and label for the item at ``ind`` index in the
+        dataset.
+        """
+        content = self._docs[ind].content
+        label = self._labels[ind]
+
+        if self._preprocess_fn is not None:
```
bwanglzu, line 60: minor thing, I think `is not None` is not necessary, when writing boolean check, python did the job for us:
```python
if foo:
      # enter True branch

if not foo:
    # enter False branch
```

finetuner/tuner/dataset/__init__.py
```diff
@@ -1,8 +1,158 @@
...
+            self._labels.append((i, 0))  # 0 is label for the anchor
+
+            num_docs += 1
+
+            for match_ind, match in enumerate(doc.matches):
+                self._locations.append((i, match_ind))
+                try:
+                    tag = match.tags['finetuner']['label']
+                except KeyError:
+                    raise KeyError(
```
bwanglzu, line 116: same as above

finetuner/tuner/dataset/__init__.py
```diff
@@ -1,8 +1,158 @@
...
+        for i, doc in enumerate(self._docs):
+            self._locations.append((i, -1))
+            self._labels.append((i, 0))  # 0 is label for the anchor
+
+            num_docs += 1
+
+            for match_ind, match in enumerate(doc.matches):
+                self._locations.append((i, match_ind))
+                try:
+                    tag = match.tags['finetuner']['label']
```
bwanglzu, line 114: this is not changed to the default key

finetuner/tuner/dataset/samplers.py
```diff
@@ -0,0 +1,220 @@
...
+        self._num_items_per_class = num_items_per_class
+        self._batch_size = batch_size
+        self._num_classes = batch_size // num_items_per_class
+
+        if len(set(labels)) < self._num_classes:
+            raise ValueError(
+                'Number of distinct classes in labels must be larger or equal than'
+                'numbers of classes in a batch'
+            )
+
```
bwanglzu, line 55: the logic here is too complex, can we assume unique `labels` we passed into the class is already denotes `num_classes`? then `num_items_per_class` can be computed by `batch_size % len(set(labels))`

bwanglzu (changes requested):
still reading the `iter` part

finetuner/tuner/dataset/__init__.py
tadejsv: No, KeyError is the correct error here, as it is the most precise

finetuner/tuner/dataset/__init__.py
tadejsv: Will fix

finetuner/tuner/dataset/samplers.py
tadejsv: No, `num_classes` and `num_items_per_class` refer to a single batch - if you have a dataset with 10000 unique labels you obviously can not fit them all into a single batch of length 64.

bwanglzu: okay

finetuner/tuner/dataset/samplers.py
```diff
@@ -0,0 +1,220 @@
...
+                cls_items = class_to_labels[cls_group][-self._num_items_per_class :]
+                del class_to_labels[cls_group][-self._num_items_per_class :]
+                batch += cls_items
+
+            batches.append(batch)
+
+        # Update length to actual one
+        self._length = len(batches)
+
+        return iter(batches)
```
bwanglzu, line 117: this iter looks very complex, can you briefly tell how it works? besides, does it means that for every time we call `next`, it will build batches again from scratch? Do you want to make this batch generation function into an independent function and implement both `__iter__` and `__next__` while in `__iter__` you yield a batch? 

This is not quite clear to me, So I left a comment

bwanglzu (commented):
left one comment regards to iter in the sampler, the rest looks good to me

finetuner/tuner/dataset/samplers.py
tadejsv: So, here's the thing: the `__iter__` methods gets called every time you do something like this
```python
for x in my_dataset: # <- This is what `__iter__` is implemented on 
    pass
```

It only gets called on the beginning of the iteration, and the `next` gets called on what is returned by `__iter__` (to be precise: here the dataset is an **iterable** and the `iter(batches)` that it returns is an **iterator**, check [python glossary](https://docs.python.org/3/glossary.html).

Now, why did I do it like this? Why not just construct `batches` in `__init__`? because in the training loop we have

```
for epoch in epochs:
    for sample in batch_sample: # Actually dataloader, but it calls sampler internally
```
And since `__iter__` will be called every epoch, with this I achieve a new shuffling of the batches every epoch.

tadejsv: This is what default pytorch samplers do as well btw

finetuner/tuner/dataset/__init__.py
```diff
@@ -1,13 +1,38 @@
...
+        """
+        Get the (preprocessed) content and label for the item at ``ind`` index in the
+        dataset.
+        """
+
+    @property
+    @abc.abstractmethod
+    def labels(self) -> List[AnyLabel]:
+        """ Get the list of labels for all items in the dataset."""
+        return self._labels
```
JoanFM, line 28: It is not abstract if u implement it

finetuner/tuner/dataset/__init__.py
```diff
@@ -1,13 +1,38 @@
...
+
+    @abc.abstractmethod
+    def __len__(self) -> int:
+        return len(self._labels)
```
JoanFM, line 32: same here

JoanFM (dismissed):

finetuner/tuner/dataset/__init__.py
```diff
@@ -1,8 +1,183 @@
...
 import numpy as np
 
 from ... import __default_tag_key__
+from ...helper import DocumentSequence
+
+
+AnyLabel = TypeVar('AnyLabel')
+
+
+class BaseDataset(abc.ABC, Generic[AnyLabel]):
```
JoanFM, line 14: Do these `Generic` really work on `Python`? or is it only like a type hint?

tadejsv: They are for type hinting yes

finetuner/tuner/dataset/__init__.py
```diff
@@ -1,8 +1,175 @@
...
+
+        if self._preprocess_fn:
+            content = self._preprocess_fn(content)
+
+        return (content, label)
+
+    @property
+    def labels(self) -> List[int]:
+        """ Get the list of integer labels for all items in the dataset."""
+        return self._labels
```
bwanglzu, line 91: why implement the same property again in child class? seems the type hints in base class and child class is different as well?

finetuner/tuner/dataset/samplers.py
bwanglzu: okay I see

bwanglzu (changes requested):
i noticed some re-implemntation of same thing in base class and child class. Also, I see it's strange to add type hints to class definition, such as `ClassDataset(BaseDataset[int])` what is the purpose?

finetuner/tuner/dataset/__init__.py
tadejsv: Regarding different type hints: this is exactly how generics work, see this for example: https://stackoverflow.com/questions/57521599/python-generics-and-subclasses

Same property is implemented again because the docstring is different.

tadejsv:
@bwanglzu the type hints added to the class that is inherited from are related to generics, see here https://docs.python.org/3/library/typing.html#generics

This is the proper way to do it, and I have checked, it actually passes mypy checking

bwanglzu (approved):
LGTM!

PR #184, opened 2021-11-02
feat(tuner): add miner for session dataset

bwanglzu:
This PR extends the current `Miner` with `Session Miners`, generate all possible tuples/triplets w.r.t each session (anchor).

### DIFF
--- a/finetuner/tuner/pytorch/miner.py
+++ b/finetuner/tuner/pytorch/miner.py
@@ -94,14 +94,12 @@ class TripletSessionMiner(BaseMiner):
         for session, group in groupby(sorted_labels_with_index, lambda x: x[0]):
             anchor_pos_session_labels = []
             anchor_pos_session_indices = []
-            neg_session_labels = []
             neg_session_indices = []
             for _, session_label, session_index in group:
                 if session_label >= 0:
                     anchor_pos_session_labels.append(session_label)
                     anchor_pos_session_indices.append(session_index)
                 else:
-                    neg_session_labels.append(session_label)
                     neg_session_indices.append(session_index)
             for anchor, pos in permutations(enumerate(anchor_pos_session_labels), 2):
                 for neg_idx in neg_session_indices:
--- a/finetuner/tuner/pytorch/miner.py
+++ b/finetuner/tuner/pytorch/miner.py
@@ -1,9 +1,27 @@
 import numpy as np
 from typing import List, Tuple
-from itertools import combinations
+from itertools import combinations, groupby
 
 from ..base import BaseMiner
 from ...helper import AnyTensor
+
+
+def _generate_all_possible_pairs(labels: List[int]):
+    return [
+        (left[0], right[0], 1) if left[1] == right[1] else (left[0], right[0], -1)
+        for left, right in combinations(enumerate(labels), 2)
+    ]
+
+
+def _generate_all_possible_triplets(labels: List[int]):
+    labels1 = np.expand_dims(labels, 1)
+    labels2 = np.expand_dims(labels, 0)
+    matches = (labels1 == labels2).astype(int)
+    diffs = matches ^ 1
+    np.fill_diagonal(matches, 0)
+    triplets = np.expand_dims(matches, 2) * np.expand_dims(diffs, 1)
+    idxes_anchor, idxes_pos, idxes_neg = np.where(triplets)
+    return list(zip(idxes_anchor, idxes_pos, idxes_neg))
 
 
 class SiameseMiner(BaseMiner):
@@ -17,10 +35,7 @@ class SiameseMiner(BaseMiner):
         :return: a pair of label indices and their label as tuple.
         """
         assert len(embeddings) == len(labels)
-        return [
-            (left[0], right[0], 1) if left[1] == right[1] else (left[0], right[0], -1)
-            for left, right in combinations(enumerate(labels), 2)
-        ]
+        return _generate_all_possible_pairs(labels)
 
 
 class TripletMiner(BaseMiner):
@@ -34,14 +49,7 @@ class TripletMiner(BaseMiner):
         :return: triplet of label indices follows the order of anchor, positive and negative.
         """
         assert len(embeddings) == len(labels)
-        labels1 = np.expand_dims(labels, 1)
-        labels2 = np.expand_dims(labels, 0)
-        matches = (labels1 == labels2).astype(int)
-        diffs = matches ^ 1
-        np.fill_diagonal(matches, 0)
-        triplets = np.expand_dims(matches, 2) * np.expand_dims(diffs, 1)
-        idxes_anchor, idxes_pos, idxes_neg = np.where(triplets)
-        return list(zip(idxes_anchor, idxes_pos, idxes_neg))
+        return _generate_all_possible_triplets(labels)
 
 
 class SiameseSessionMiner(BaseMiner):
@@ -52,9 +60,19 @@ class SiameseSessionMiner(BaseMiner):
 
         :param embeddings: embeddings from model, should be a list of Tensor objects.
         :param labels: labels of each embeddings, consist of session id and label.
+          the labels are either -1 or 1.
         :return: a pair of label indices and their label as tuple.
         """
+        rv = []
         assert len(embeddings) == len(labels)
+        for session, group in groupby(
+            sorted(labels, key=lambda x: x[0]), lambda x: x[0]
+        ):
+            session_labels = [item for _, item in group]
+            rv.extend(
+                _generate_all_possible_pairs(session_labels)
+            )  # all possible tuples in the session
+        return rv
 
 
 class TripletSessionMiner(BaseMiner):
@@ -67,4 +85,11 @@ class TripletSessionMiner(BaseMiner):
         :param labels: labels of each embeddings, consist of session id and label.
         :return: triplet of label indices follows the order of anchor, positive and negative.
         """
-        pass
+        rv = []
+        assert len(embeddings) == len(labels)
+        for session, group in groupby(
+            sorted(labels, key=lambda x: x[0]), lambda x: x[0]
+        ):
+            session_labels = [item for _, item in group]
+            rv.extend(_generate_all_possible_triplets(session_labels))
+        return rv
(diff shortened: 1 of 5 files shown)


finetuner/tuner/pytorch/miner.py
```diff
@@ -63,15 +63,21 @@ def mine(
...
         :return: a pair of label indices and their label as tuple.
         """
-        rv = []
         assert len(embeddings) == len(labels)
+        rv = []
+        labels_with_index = [item + (index,) for index, item in enumerate(labels)]
+        sorted_labels_with_index = sorted(labels_with_index, key=lambda x: x[0])
         for session, group in groupby(
-            sorted(labels, key=lambda x: x[0]), lambda x: x[0]
+            sorted(sorted_labels_with_index, key=lambda x: x[0]), lambda x: x[0]
```
JoanFM: sorted is already sorted right?

bwanglzu: refresh :)

tests/unit/tuner/torch/test_miner.py
```diff
@@ -27,6 +42,11 @@ def labels():
     return [1, 3, 1, 3, 2, 4, 2, 4]
 
 
+@pytest.fixture
+def session_labels():
+    return [(1, 1), (2, -1), (2, 1), (1, -1), (3, -1), (3, 1), (3, -1), (3, -1)]
```
tadejsv, line 47: You need to take into account that each session has an anchor with label `0`

tests/unit/tuner/torch/test_miner.py
```diff
@@ -75,3 +95,60 @@ def test_triplet_miner_given_insufficient_inputs(
...
+    for item in rv:
+        idx_left, idx_right, label = item
+        # find corresponded label idx
+        label_left = session_labels[idx_left]
+        label_right = session_labels[idx_right]
+        if label_left == label_right:
+            expected_label = 1
+        else:
+            expected_label = -1
+        assert label == expected_label
```
tadejsv, line 114: Here I would prefer if you explicitly list the tuples you expect to be produced (since it's only 8 of them, it's feasible) or produce them by a "reliable" method.

The way test is written now does not pin down the result enough in my opinion

tests/unit/tuner/torch/test_miner.py
```diff
@@ -75,3 +95,60 @@ def test_triplet_miner_given_insufficient_inputs(
...
+def test_siamese_session_miner_given_insufficient_inputs(
+    embeddings, session_labels, siamese_session_miner, cut_index
+):
+    embeddings = embeddings[:cut_index]
+    session_labels = session_labels[:cut_index]
+    rv = list(siamese_session_miner.mine(embeddings, session_labels))
+    assert len(rv) == 0
+
+
+def test_triplet_session_miner(embeddings, session_labels, triplet_session_miner):
```
tadejsv, line 127: Same for this test, list expected result explicitly.

finetuner/tuner/pytorch/miner.py
```diff
@@ -34,11 +49,61 @@ def mine(
...
+        sorted_labels_with_index = sorted(labels_with_index, key=lambda x: x[0])
+        for session, group in groupby(sorted_labels_with_index, lambda x: x[0]):
+            _, session_labels, session_indices = zip(*group)
+            pairs = _generate_all_possible_pairs(session_labels)
+            rv.extend(
+                [
+                    (session_indices[left_idx], session_indices[right_idx], label)
+                    for left_idx, right_idx, label in pairs
+                ]
+            )
```
tadejsv, line 78: Please, this should be a for loop. Unless you can show we get significant time savings (not microseconds per batch), readability has absolute priority

finetuner/tuner/pytorch/miner.py
```diff
@@ -34,11 +49,61 @@ def mine(
...
+            _, session_labels, session_indices = zip(*group)
+            triplets = _generate_all_possible_triplets(session_labels)
+            rv.extend(
+                [
+                    (
+                        session_indices[left_idx],
+                        session_indices[middle_idx],
+                        session_indices[right_idx],
+                    )
+                    for left_idx, middle_idx, right_idx in triplets
```
tadejsv, line 106: Same - for loop unless you can demonstrate significant time savings

finetuner/tuner/pytorch/miner.py
```diff
@@ -34,11 +49,61 @@ def mine(
...
+                    (session_indices[left_idx], session_indices[right_idx], label)
+                    for left_idx, right_idx, label in pairs
+                ]
+            )
+        return rv
+
+
+class TripletSessionMiner(BaseMiner):
+    def mine(
+        self, embeddings: List[AnyTensor], labels: List[Tuple[int, int]]
```
tadejsv, line 84:
```suggestion
        self, embeddings: AnyTensor, labels: List[Tuple[int, int]]
```

finetuner/tuner/pytorch/miner.py
```diff
@@ -34,11 +49,61 @@ def mine(
...
-        np.fill_diagonal(matches, 0)
-        triplets = np.expand_dims(matches, 2) * np.expand_dims(diffs, 1)
-        idxes_anchor, idxes_pos, idxes_neg = np.where(triplets)
-        return list(zip(idxes_anchor, idxes_pos, idxes_neg))
+        return _generate_all_possible_triplets(labels)
+
+
+class SiameseSessionMiner(BaseMiner):
+    def mine(
+        self, embeddings: List[AnyTensor], labels: List[Tuple[int, int]]
```
tadejsv, line 57:
```suggestion
        self, embeddings: AnyTensor, labels: List[Tuple[int, int]]
```

It's not a list, simply a tensor with first dimension `N`

finetuner/tuner/pytorch/miner.py
```diff
@@ -34,11 +49,61 @@ def mine(
...
+    ) -> List[Tuple[int, ...]]:
```
tadejsv, line 58:
```suggestion
    ) -> List[Tuple[int, int, int]]:
```

finetuner/tuner/pytorch/miner.py
```diff
@@ -34,11 +49,61 @@ def mine(
...
+                    for left_idx, right_idx, label in pairs
+                ]
+            )
+        return rv
+
+
+class TripletSessionMiner(BaseMiner):
+    def mine(
+        self, embeddings: List[AnyTensor], labels: List[Tuple[int, int]]
+    ) -> List[Tuple[int, ...]]:
```
tadejsv, line 85:
```suggestion
    ) -> List[Tuple[int, int, int]]:
```

finetuner/tuner/pytorch/miner.py
```diff
@@ -34,11 +49,61 @@ def mine(
...
+        """Generate triplets from input embeddings and labels.
+
+        :param embeddings: embeddings from model, should be a list of Tensor objects.
+        :param labels: labels of each embeddings, consist of session id and label.
+        :return: triplet of label indices follows the order of anchor, positive and negative.
+        """
+        assert len(embeddings) == len(labels)
+        rv = []
+        labels_with_index = [item + (index,) for index, item in enumerate(labels)]
+        sorted_labels_with_index = sorted(labels_with_index, key=lambda x: x[0])
```
tadejsv, line 95: Can I ask why are we sorting? We are doing `groupby` later anyway

finetuner/tuner/pytorch/miner.py
```diff
@@ -34,11 +49,61 @@ def mine(
...
+        for session, group in groupby(sorted_labels_with_index, lambda x: x[0]):
+            _, session_labels, session_indices = zip(*group)
+            triplets = _generate_all_possible_triplets(session_labels)
```
tadejsv, line 98: This implementation is actually wrong. You can not assume that items with label `-1` are similar among each other. I can also see that from your test, because for session `3` you should also get 0 triplets as it is now, but you say you get 6

tadejsv (changes requested):

finetuner/tuner/pytorch/miner.py
bwanglzu: if you [read](https://stackoverflow.com/a/8116674), groupby only works on sorted data, it's stupid but true.

finetuner/tuner/pytorch/miner.py
bwanglzu: i see, you're right

finetuner/tuner/pytorch/miner.py
```diff
@@ -42,3 +42,71 @@ def mine(
...
+        rv = []
+        labels_with_index = [item + (index,) for index, item in enumerate(labels)]
+        sorted_labels_with_index = sorted(labels_with_index, key=lambda x: x[0])
+        for session, group in groupby(sorted_labels_with_index, lambda x: x[0]):
+            _, session_labels, session_indices = zip(*group)
+            session_pairs = []
+            for left, right in combinations(enumerate(session_labels), 2):
+                if left[1] >= 0 and right[1] >= 0:
+                    # 0 represents for anchor, 1 for positive, they form positive pairs.
+                    session_pairs.append((left[0], right[0], 1))
```
tadejsv, line 68: I would append to `rv` here already, so we don't need to do another loop

finetuner/tuner/pytorch/miner.py
```diff
@@ -42,3 +42,71 @@ def mine(
...
+          anchor, positive, negative are 0, 1 and -1.
+        :return: triplet of label indices follows the order of anchor, positive and negative.
+        """
+        assert len(embeddings) == len(labels)
+        rv = []
+        labels_with_index = [item + (index,) for index, item in enumerate(labels)]
+        sorted_labels_with_index = sorted(labels_with_index, key=lambda x: x[0])
+        for session, group in groupby(sorted_labels_with_index, lambda x: x[0]):
+            _, session_labels, session_indices = zip(*group)
+            for left, middle, right in permutations(enumerate(session_labels), 3):
```
tadejsv, line 101: I'm thinking we can make this more efficient. Currently most of permutations we loop over will get discarded.

We already know that only positive and anchors can be at first two places, and negative has to be at the last place. So what you could do is to:
- first, get a list of positive/anchor indices
- get a list of negative indices
- loop over permutations of size 2 for pos/anchor
- inside this loop, loop over negative indices
- add each such combination to result

tadejsv (approved):
Looks good!

bwanglzu:
note: i fixed tf version in this PR. this needs to be investigated in the future.

PR #188, opened 2021-11-04
ci: pin framework version

bwanglzu:
~once after framework version settled will try to increase tf version.~

Note: I also decreased `num_total` in CI integration test `fit` to make it a bit faster.

### DIFF
--- a/.github/requirements-cicd-gpu.txt
+++ b/.github/requirements-cicd-gpu.txt
@@ -1,5 +1,5 @@
 https://github.com/jina-ai/jina/archive/refs/heads/master.zip
-tensorflow==2.5.0
+tensorflow<=2.6.1
 paddlepaddle-gpu<=2.2.0
 torch<=1.9.0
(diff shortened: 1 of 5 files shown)


.github/requirements-cicd-gpu.txt
```diff
@@ -1,6 +1,6 @@
 https://github.com/jina-ai/jina/archive/refs/heads/master.zip
-tensorflow==2.5.0
-paddlepaddle-gpu
-torch
-torchvision
-scipy
\ No newline at end of file
+tensorflow<=2.5.2
```
tadejsv, line 2: I would still pin the exact version. What is the advantage of doing `<=` instead of `==`?

tadejsv: And maybe we can pin it to 2.6.1, and then pin `keras` as well - that would prevent the issue we had

tadejsv (approved):

PR #192, opened 2021-11-08
feat: reformat pipeline

tadejsv:
Todo

- [x] general cleanup, fix GPU + integration tests
- [x] test for transformers model
- [x] rework toydata
- [x] ressurect `eval`
- [x] pass optimizer as class

- [x] bubble up new args to `fit` functions
- [ ] update documentation: update tuner docs, replace LSTM example with transformers
- [x] fix all GPU tests
- [x] fix toydata tests
- [x] port to paddle
- [x] port to keras
    - refactor loss and miner @bwanglzu 
    - make sure integration and general tests match torch
- [x] make sure it works with labeler
    - Requires fixing #206


finetuner/tuner/pytorch/__init__.py
```diff
@@ -39,124 +51,136 @@ def _get_optimizer(
...
+                momentum=optimizer_kwargs["momentum"],
+                nesterov=optimizer_kwargs["nesterov"],
             )
 
     def _eval(
         self,
-        data: AnyDataLoader,
-        description: str = 'Evaluating',
+        data: DataLoader,
+        description: str = "Evaluating",
```
hanxiao, line 81: single quote

finetuner/tuner/pytorch/__init__.py
```diff
@@ -39,124 +51,136 @@ def _get_optimizer(
...
+                inputs = _to_device(inputs, self.device)
+                labels = _to_device(labels, self.device)
 
-                with torch.inference_mode():
-                    embeddings = [self._embed_model(inpt) for inpt in inputs]
-                    loss = self._loss(embeddings, label)
+                # Can not use inference mode due to (when having too many triplets)
+                # https://github.com/pytorch/pytorch/issues/60539
+                with torch.no_grad():
+                    embeddings = self.embed_model(inputs)
```
hanxiao, line 102: https://github.com/jina-ai/finetuner/blob/main/finetuner/embedding.py

finetuner/tuner/pytorch/__init__.py
```diff
@@ -171,34 +195,49 @@ def fit(
...
+                eval_dataset, batch_size, num_items_per_class, False
+            )
+            eval_dl = DataLoader(
+                dataset=eval_dataset,
+                batch_sampler=eval_batch_sampler,
+                collate_fn=collate_fn,
+            )
+
+        # Get miner
+        self._miner = miner if miner else self._loss.get_default_miner(train_dataset)
```
JoanFM, line 220:
```suggestion
        self._miner = miner or self._loss.get_default_miner(train_dataset)
```

finetuner/tuner/pytorch/datasets.py
```diff
@@ -1,12 +1,11 @@
...
+from ..dataset import ClassDataset, SessionDataset
 
 
-class SiameseDataset(SiameseMixin, BaseDataset, IterableDataset):
+class PytorchClassDataset(ClassDataset, Dataset):
     ...
 
 
-class TripletDataset(TripletMixin, BaseDataset, IterableDataset):
+class PytorchSessionDataset(SessionDataset, Dataset):
```
JoanFM, line 10: why is this neederd? Why is `DataSet` and `SessionDataSet` orthogonal and needed to be mixed in?

JoanFM: It seems like a big smell

finetuner/tuner/__init__.py
```diff
@@ -46,10 +47,8 @@ def fit(
     :param batch_size: The batch size to use for training and evaluation
     :param loss: Which loss to use in training. Supported
         losses are:
-        - ``CosineSiameseLoss`` for Siamese network with cosine distance
-        - ``EuclideanSiameseLoss`` for Siamese network with eculidean distance
-        - ``CosineTripletLoss`` for Triplet network with cosine distance
-        - ``EuclideanTripletLoss`` for Triplet network with eculidean distance
+        - ``SiameseLoss`` for Siamese network with cosine distance
```
JoanFM, line 50: losses and distance are not separated?

finetuner/tuner/pytorch/losses.py
```diff
@@ -1,143 +1,135 @@
...
-    and 0 otheriwse. The ``dist`` is the euclidean distance between the embeddings of
-    the objects, and ``margin`` is some number, used here to ensure better stability
-    of training.
-
-    The final loss is the average over losses for all pairs of objects in the batch.
+    The final loss is the average over losses for all pairs given by the indices.
     """
 
-    arity = 2
+    def __init__(self, distance: str = "cosine", margin: float = 1.0):
```
JoanFM, line 41: single quote

finetuner/tuner/pytorch/losses.py
```diff
@@ -1,143 +1,135 @@
...
-        is_similar = (target > 0).to(torch.float32)
+        ind_one, ind_two, target = indices
+        dist_matrix = get_distance(embeddings, self.distance)
+        dists = dist_matrix[ind_one, ind_two]
+        target = target.to(torch.float32)
 
-        loss = 0.5 * torch.square(
-            is_similar * eucl_dist + (1 - is_similar) * F.relu(self.margin - eucl_dist)
-        )
+        loss = target * dists + (1 - target) * F.relu(self.margin - dists)
```
JoanFM, line 70: Why do not use the ContrastiveLoss implementation from `pytorch`?

JoanFM (changes requested):

finetuner/tuner/pytorch/datasets.py
tadejsv: This is due to pytorch DataLoader. It requires the dataset input to it to subclass its `Dataset`, see [here](https://pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader).

It actually works without that (because `Dataset` in essence only implements `__getitem__`), but that is not documented, so I want to properly subclass `Dataset` to avoid any "suprises" in the future

finetuner/tuner/__init__.py
tadejsv: I've refactored this so that they are separated, you can create a loss object with specified distance

finetuner/tuner/pytorch/losses.py
tadejsv: Where do you see ContrastiveLoss in Pytorch?

finetuner/tuner/pytorch/datasets.py
```diff
@@ -1,12 +1,11 @@
-from torch.utils.data import IterableDataset
+from torch.utils.data import Dataset
```
JoanFM, line 1:
```suggestion
from torch.utils.data import Dataset as PytorchDataSet
```

finetuner/tuner/pytorch/datasets.py
```diff
@@ -1,12 +1,11 @@
...
 
-from ..base import BaseDataset
-from ..dataset import SiameseMixin, TripletMixin
+from ..dataset import ClassDataset, SessionDataset
 
 
-class SiameseDataset(SiameseMixin, BaseDataset, IterableDataset):
+class PytorchClassDataset(ClassDataset, Dataset):
```
JoanFM, line 6:
```suggestion
class PytorchClassDataset(ClassDataset, PytorchDataset):
```

finetuner/tuner/pytorch/datasets.py
JoanFM:
```suggestion
class PytorchSessionDataset(SessionDataset, PytorchDataset):
```

JoanFM (commented):
I agree, I would rename the class for less confusion

finetuner/tuner/__init__.py
JoanFM: Correct comment then

finetuner/tuner/base.py
```diff
@@ -1,40 +1,107 @@
...
+            respectively
+        :param distances: A tensor matrix of pairwise distance between each two item
+            embeddings
+
+        :return: A tuple of 1D tensors, denoting indices or properties of returned
+            tuples
+        """
 
 
 class BaseLoss:
```
JoanFM, line 62: I think this designs does not impose any abstract method to BaseLoss, it seems that to be a valid loss u just need to provide a way to get default miner, but We should add a method `compute` or something, that in the `Pytorch` it should go to `forward`. Or maybe a `__call__` interface, but this BaseLoss is a poor interface

JoanFM (changes requested):

finetuner/tuner/base.py
tadejsv: Ok, let me think about it. using `__call__` won't work because then we mess with pytorch's implementation

bwanglzu:
can we already start review? I see a lot of test failing.

tadejsv:
@bwanglzu yes please, start reviewing. Test are and will be failing probably until all other frameworks are done, so don't wait for that. Currently I just make sure that localy right tests pass

finetuner/tuner/base.py
```diff
@@ -1,112 +1,138 @@
...
+LabelType = TypeVar('LabelType')
+
+
+class BaseMiner(abc.ABC, Generic[AnyTensor, LabelType]):
+    @abc.abstractmethod
+    def mine(self, labels: LabelType, distances: AnyTensor) -> Tuple[AnyTensor, ...]:
+        """Generate mined tuples from labels and item distances.
+
+        :param labels: labels of each item
+        :param distances: A tensor matrix of pairwise distance between each two item
```
bwanglzu, line 20: now sampler give distances rather than embeddings?

finetuner/__init__.py
```diff
@@ -6,25 +6,29 @@
...
 def fit(
     model: 'AnyDNN',  #: must be an embedding model
-    train_data: 'DocumentArrayLike',
-    eval_data: Optional['DocumentArrayLike'] = None,
+    train_data: 'DocumentSequence',
+    eval_data: Optional['DocumentSequence'] = None,
+    preprocess_fn: Optional[Callable] = None,
+    collate_fn: Optional[Callable] = None,
     epochs: int = 10,
+    miner: Optional['BaseMiner'] = None,
```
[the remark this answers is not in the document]
bwanglzu: i see you've created `get_default_miner` below, then we can really hidden this parameter.
Another thing is we parse a loss `SiameseLoss` as default, how can we allow user to change `distance_metric`? such as euclidean?

bwanglzu (changes requested):

finetuner/__init__.py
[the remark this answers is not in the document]
tadejsv: > then we can really hidden this parameter.

what parameter

> how can we allow user to change distance_metric?

User can pass loss as an object, instead of string

finetuner/tuner/pytorch/__init__.py
```diff
@@ -1,71 +1,56 @@
...
 from ..summary import ScalarSequence, Summary
-from ...helper import DocumentArrayLike, AnyDataLoader
+from . import losses
+from .datasets import PytorchClassDataset, PytorchSessionDataset
 
 
-class PytorchTuner(BaseTuner):
-    def _get_loss(self, loss: Union[BaseLoss, str]) -> BaseLoss:
+def _to_device(
+    inputs: Union[torch.Tensor, Mapping[str, torch.Tensor]], device: torch.device
```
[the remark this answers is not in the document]
tadejsv: Where do you see `Sequence` here?

finetuner/tuner/pytorch/__init__.py
```diff
@@ -1,71 +1,56 @@
...
-        return DataLoader(
-            dataset=ds(inputs=inputs),
-            batch_size=batch_size,
-            shuffle=shuffle,
-        )
+    def _get_dataset(
+        self, data: DocumentSequence, preprocess_fn: Optional[Callable]
+    ) -> Union[PytorchClassDataset, PytorchSessionDataset]:
+        if __default_tag_key__ in data[0].tags:
+            dataset = PytorchClassDataset(data, preprocess_fn=preprocess_fn)
```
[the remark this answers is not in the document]
tadejsv: No, logic is like this:
- root document has no `__default_tag_key__` -> this is session dataset, as then matches should have this property
- root document has ``__default_tag_key__` -> this is class dataset

finetuner/tuner/base.py
tadejsv: Yes, passing embeddings was inefficient - if there were a million triplets, it would have to compute million distances, instead of taking a million numbers from distance matrix. In almost all cases there are more triplets/pairs than elements in distance matrix, so this is much more efficient

finetuner/tuner/pytorch/__init__.py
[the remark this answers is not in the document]
tadejsv: Ah, I see what you mean. I'll add it, thanks

finetuner/tuner/base.py
tadejsv: @JoanFM  I've added a `compute` function to the base loss class

finetuner/__init__.py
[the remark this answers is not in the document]
tadejsv: @bwanglzu miner is now part of the loss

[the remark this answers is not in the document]
bwanglzu: i mean do we need to expose `miner` to `fit` function? in the end, miner depends on two things:
1. if user already provides pos/neg.
2. if it is a siamese or triplet.

then in the back we can initialise different miner instances automatically, rather than let user pass a miner class instance to the fit fucntion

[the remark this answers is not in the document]
tadejsv: No, ability for the user to set the miner is necessary. This is because in the future we will have different types of negative miners, which the user would have to explicitly select.

finetuner/tuner/pytorch/__init__.py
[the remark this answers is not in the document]
bwanglzu: OKAY

finetuner/labeler/__init__.py
```diff
@@ -39,6 +39,10 @@ def fit(
     dam_path = tempfile.mkdtemp()
     stop_event = threading.Event()
 
+    # Remove all labels, as they are not needed
+    for doc in train_data:
```
JoanFM, line 43: Then the argument `clear_labels_on_start` is not used?

finetuner/tuner/base.py
```diff
@@ -1,112 +1,151 @@
...
+            embeddings
+
+        :return: A tuple of 1D tensors, denoting indices or properties of returned
+            tuples
+        """
+
+
+class BaseClassMiner(BaseMiner[AnyTensor, AnyTensor], Generic[AnyTensor]):
+    @abc.abstractmethod
+    def mine(self, labels: AnyTensor, distances: AnyTensor) -> Tuple[AnyTensor, ...]:
```
JoanFM, line 29: are we sure this is the wanted interface? isn't embeddings + dist easier to grasp?

JoanFM, line 29: are we sure this is the wanted interface? isn't embeddings + dist easier to grasp?

finetuner/tuner/dataset/__init__.py
```diff
@@ -138,7 +137,7 @@ def __init__(
                         ' When using ClassDataset all documents need this tag'
                     )
 
-                self._labels.append((i, tag))
+                self._labels.append((i, int(tag)))
```
JoanFM, line 140: why is this conversion needed? this is quite an assumption right?

finetuner/tuner/dataset/samplers.py
```diff
@@ -63,12 +63,20 @@ def __init__(
...
 
-        # Approximate (max) length
-        self._length = ceil(
-            sum([x[1] for x in self._cls_group_counts]) / self._num_classes
-        )
+        # Prepare the current batches
+        self._prepare_batches()
 
     def __iter__(self) -> Iterator[Tuple[int, ...]]:
+        batches = self._batches
```
JoanFM, line 70: this code seems weird.

Why do I get `batches` in advance before a call that changes `self._batches` to return the old ones?

finetuner/tuner/dataset/samplers.py
```diff
@@ -63,12 +63,20 @@ def __init__(
...
+
+        # Prepare next batches - the reason they are prepared in advance is so that
+        # len() gives the correct length
+        self._prepare_batches()
```
JoanFM, line 74: I think the proper implementation is to have some lazy property for `length`. This is quite a smell

finetuner/tuner/base.py
tadejsv: Labels are needed, because they tell you which items are similar or not. The distances are here because they provide all the information needed from the embeddings (I don't know of miners that would need more information), and it is more efficient not to re-compute them every time

finetuner/tuner/keras/miner.py
```diff
@@ -0,0 +1,153 @@
+from itertools import combinations, groupby, permutations
+from typing import Tuple
+
+import tensorflow as tf
+
+from ..base import BaseClassMiner, BaseSessionMiner
+
+
+class SiameseMiner(BaseClassMiner[tf.Tensor]):
```
JoanFM, line 9: I still think that the Miner design is weird, I do not understand why it has to be tight to a `framework`. 

It should be depeding on labels, any embeddings and distances callables

JoanFM (commented):
I am not sure about the miner design

finetuner/labeler/__init__.py
tadejsv: Oh, haven't seen that one - will fix, thanks

finetuner/tuner/keras/miner.py
tadejsv: Computing triplets is significantly faster for large batches when using GPU - which requires the use of tensors.

finetuner/tuner/dataset/samplers.py
tadejsv: If you call `__len__` before calling `__iter__` you will get the wrong length if `self._batches` are not precomputed. So the only solution is to pre-compute the next `_batches` when you call `__iter__`. Because of random sampling `__len__` can not easily be computed

finetuner/tuner/dataset/__init__.py
tadejsv: I am converting because it may be a float (the allowed values are still only `-1` and `1`)

finetuner/tuner/keras/miner.py
tadejsv: To expand on this - for class dataset this is a problem with n^3 time complexity, which is highly parallelizable. If `n=512` I got very significant time savings. Keeping everything on CPU/numpy would hinder the performance when training on GPU

finetuner/tuner/dataset/samplers.py
tadejsv: I have now reformated this, so that the sampler is also the iterator, and new batches are only computed once the current ones are exhausted.

finetuner/tuner/keras/miner.py
JoanFM: The usage of tensors and the abstractions I do not see why they are related.

JoanFM: My concern is the explosion of work to add a new mining strategy, for instance some hard mining logic, I will have to implement it for each framework. Beyond, it surprises me that computing the triplets becoming the bottleneck of the system.

finetuner/labeler/__init__.py
```diff
@@ -27,25 +30,39 @@ def fit(
...
     """
     dam_path = tempfile.mkdtemp()
     stop_event = threading.Event()
 
+    # Remove all labels, as they are not needed
+    for doc in train_data:
+        try:
+            del doc.tags[__default_tag_key__]
+        except:
+            pass
```
bwanglzu, line 51: i would prefer a bit more defensive, not a fan of pass an exception sliently
```suggestion
    for doc in train_data:
        if doc.tags.get(__default_tag_key__):
            del doc.tags[__default_tag_key__]
```

bwanglzu (approved):
do not have further comments, seems we still have some docstring missing. but we can optimize it later.

finetuner/labeler/__init__.py
tadejsv: Check that your suggestions acutally work - the second one gives the error
```
  File "/home/tadej/miniconda3/envs/finetuner/lib/python3.9/site-packages/jina/types/mixin.py", line 97, in pop
    self._pb_body.ClearField(k)
ValueError: Protocol message has no "finetuner" field.
```

bwanglzu (approved):

PR #196, opened 2021-11-09
fix(tailor): fix to emebdding model

bwanglzu:
bug for `to_embedding_layer` (fixed for torch and paddle, tf works fine):

Reason: after attach a new layer to the end of model (user specify a different `output_dim`), `Paddle` and `Pytorch` added `_LinearAtLast` module, and `linear` layer will be considered as a part of `_LinearAtLast`, `is_embedding_layer` became `False`.

before: 
image

after:

image


maximilianwerk:
LGTM

maximilianwerk:
Could you add a test for it?

PR #204, opened 2021-11-15
docs: siamese networks

florian-hoenicke:


README.md
```diff
@@ -22,8 +22,8 @@ applications.
 
 🎛 **Designed for finetuning**: a human-in-the-loop deep learning tool for leveling up your pretrained models in domain-specific neural search applications.
 
-🔱 **Powerful yet intuitive**: all you need is `finetuner.fit()` - a one-liner that unlocks rich features such as
-siamese/triplet network, interactive labeling, layer pruning, weights freezing, dimensionality reduction.
+🔱 **Powerful yet intuitive**: all you need is `finetuner.fit()` - a one-liner that unlocks rich features such as:
+training your network as siamese network using contrastive loss like Triplet Loss, interactive labeling, layer pruning, weights freezing and dimensionality reduction.
```
JoanFM, line 26: are we sure TripletLoss is a term used by frameworks? I think Triplet and Siamese refer to architectures, while Losses can differ.

tadejsv: Loss and framework are actually the same in our setting, so this change is wrong

florian-hoenicke: Triplet loss is one contrastive loss or also called ranking loss. Explained here: https://gombru.github.io/2019/04/03/ranking_loss/
Siamese network means that you have shared layers - has nothing to do with the loss function.
You can also train a Siamese network on classification loss.

JoanFM: Yes, what I mean that are we sure that losses are named after TripletLoss?

tadejsv: In **our** setting, loss and network, when it comes to Triplet/Siamese mean the same thing. 

The terms are not precisely specified in the literature - e.g. wikipedia states that siamese network works on two inputs (and thus precludes triplet loss), and the term triplet network is also used, see for example https://arxiv.org/pdf/1412.6622.pdf

florian-hoenicke: Wow, then the wikipedia article is contradictory itself.
It also says:
`Learning in twin networks can be done with triplet loss or contrastive loss.`
This is also weird since triplet loss is one of the contrastive losses. I guess I should have not argued with a wiki article in the first place :D 
But recent machine learning papers call it triplet loss and everything which has shared weights siamese network.
https://openaccess.thecvf.com/content_ECCV_2018/html/Xingping_Dong_Triplet_Loss_with_ECCV_2018_paper.html
https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=9082602
https://ieeexplore.ieee.org/abstract/document/9207667
http://www.dei.unipd.it/~ferro/CLEF-WN-Drafts/CLEF2020/paper_143.pdf
https://www.sciencedirect.com/science/article/abs/pii/S0045790620306108

hanxiao (changes requested):
reject. it doesn't fit our implementation

PR #207, opened 2021-11-16
fix: make get_framework robust

tadejsv:
Right now, `get_framework` is not robust, it fails against any kind of third party modules (ones created by user or from libraries like `transformers`). See this test below (fails with current `main`):

```python
def test_torch_custom():
    class MyModel(torch.nn.Module):
        pass

    model = MyModel()
    assert 'torch' == get_framework(model)
```

This PR fixes that by using `isinstance` for checking. However, this requires importing the corresponding framework (e.g. `torch`) - which can be a problem, as it would require all 3 frameworks to be installed just for inference. To circumvent this, a check is performed whether the framework is installed at all.

This is not ideal still - namely, if the user has all 3 frameworks installed (very rare case, but still), it can be the case that all 3 are imported just to perform this check. This is a disadvantage of what we are doing now - inference, instead of giving user the choice via exposing properly framework-specific class objects.

### DIFF
--- a/finetuner/helper.py
+++ b/finetuner/helper.py
@@ -1,3 +1,4 @@
+import importlib.util
 from typing import (
     TypeVar,
     Sequence,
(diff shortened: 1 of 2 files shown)


finetuner/helper.py
```diff
@@ -52,15 +53,30 @@ def get_framework(dnn_model: AnyDNN) -> str:
...
+        import torch
+
+        if isinstance(dnn_model, torch.nn.Module):
+            framework = 'torch'
+    if framework is None and importlib.util.find_spec('torch'):
+        import paddle
+
+        if isinstance(dnn_model, paddle.nn.Layer):
+            framework = 'paddle'
+    if framework is None and importlib.util.find_spec('torch'):
```
bwanglzu, line 69: this is tf.

finetuner/helper.py
```diff
@@ -52,15 +53,30 @@ def get_framework(dnn_model: AnyDNN) -> str:
...
-    else:
+
+    framework = None
+
+    if importlib.util.find_spec('torch'):
+        import torch
+
+        if isinstance(dnn_model, torch.nn.Module):
+            framework = 'torch'
+    if framework is None and importlib.util.find_spec('torch'):
```
bwanglzu, line 64: paddle

bwanglzu (changes requested):
seems you're inspect torch in every if else branch? how did the test pass?

finetuner/helper.py
tadejsv: Ah thanks.... as for how tests pass - we have all frameworks installed. No way to test this properly unless we have an environment with only one framework- but that's too much effort for only 1 test function

tadejsv: Fixed now @bwanglzu

bwanglzu (approved):
LGTM!

PR #209, opened 2021-11-17
fix: raise value error in loss if get nothing from miner

bwanglzu:


tests/unit/tuner/paddle/test_loss.py
```diff
@@ -60,3 +60,26 @@ def test_compute(loss_cls, indices, exp_result):
...
+        loss_cls(distance='euclidean').compute(embeddings, indices)
+
+
+@pytest.mark.gpu
+@pytest.mark.parametrize(
+    'loss_cls',
+    [SiameseLoss, TripletLoss],
+)
+def test_compute_loss_given_insufficient_data_gpu(loss_cls):
+    indices = [paddle.to_tensor([]) for _ in range(3)]
```
tadejsv, line 82: Make sure these tests are actually done on the GPU - set the device for tensors. Same goes for other frameworks

tadejsv (changes requested):

gmastrapas (approved):

PR #215, opened 2021-11-21
refactor(dataset): make preprocess_fn work on document

hanxiao:


tadejsv:
@hanxiao Why? I implemented it the way I did so that you can avoid as much Document overhead as possible, and get data into native format as soon as possible. What's the point of using the document here really?

hanxiao:
because you can leverage Document API as documented here 
image

Also, here is a real example how to leverage it in practice:

http://54.93.57.58:8888/notebooks/han/totally_looks_like_v3.ipynb#!

image

Note that, directly working with `.content` is not appreciated and it was part of Jina 1.0 design.

tadejsv:
@hanxiao With your change it is now much trickier to use already existing augmentation packages, like torchvision or albumentations. 

I mean, you can still use them, but in a silly way - instead of using them to directly convert to a tensor, as you do usually with this packages, you have to extract the content (blob) of the document, transform it, and then write it back to document - for no reason, as it will be converted to tensor anyway right after. I am not sure how much overhead this adds (some for sure), but just this added complexity would make me, as a user, think twice about using finetuner.

And the benefit is very small - all we are doing here is re-inventing the wheel - the transformations you do in the example above with jina are standard for the augmentation packages I mentioned.

Also, I think you broke the example of preprocessing that I put in the "Data Format" part of the documentation with this PR

hanxiao:
Then please think a way to combine both.

PR #222, opened 2021-11-22
fix(docbot): feedback tooltip ui style

nomagick:
This fixes the feedback tooltip broken issue when the answer is too short.
Although finetuner docbot rarely return short answers to get a buggy screenshot,
It's better to keep code the same with Jina doc.

From:
CleanShot 2021-11-22 at 22 53 48@2x

To:
CleanShot 2021-11-23 at 00 59 52@2x


maximilianwerk (approved):
LGTM

PR #223, opened 2021-11-23
feat(evaluation): add evaluation metrics

### LINKED ISSUE #203: Evaluation functions
Have an evaluation function with the following interface

```
def get_evaluation(queries: DocumentSequence, catalog: Union[DocumentArray, DocumentArrayMemmap], model: AnyDNN, metrics: List[str] = None):
```

where
- `queries` are the to-be-scored Documents with either positive results as matches or class informations,
- `catalog` are the potential results,
- `model` is any model and
- `metrics` the names of the metrics, where `None` means all.

The evaluation function should compute the embeddings and compute/output requested metrics.

Furthermore, we should have a demo evaluation for the FMNIST dataset as an integration test (potentially sub-sampled for faster execution).

gmastrapas:
Closes #203 

Implement evaluation metrics in `finetuner.tuner.evaluation` as wrappers over the metrics at `jina.math.evaluation`


tests/unit/tuner/test_evaluation.py
```diff
@@ -0,0 +1,164 @@
...
+# doc with no matches
+doc7 = Document(matches=[], tags={__default_tag_key__: {"targets": {"1": 3, "3": 1}}})
+
+# doc with no ground truth
+doc8 = Document(
+    matches=[doc1, doc2, doc3, doc4], tags={__default_tag_key__: {"targets": {}}}
+)
+
+# doc with no matches and no ground truth
+doc9 = Document(matches=[], tags={__default_tag_key__: {"targets": {}}})
```
maximilianwerk, line 32: This should all be fixtures. Otherwise, one test changing one of the docs would have side-effect towards other tests. Having them as fixtures makes sure, tests are completely decoupled. E.g.
```
@pytest.fixture
def doc5():
    return Document(matches=[Document(id="1")]...)

def test_r_precision(doc5, ...):
    ...
```

maximilianwerk (changes requested):
I like it. Some comments regarding the tests. 

Furthermore, I am not 100% sure is whether we should have a lot of static methods or rather put all these methods in a separate module in order to keep the class simpler and isolate functionality better. Let's keep it like you've done it but move it out in case the class gets more logic.

tests/unit/tuner/test_evaluation.py
gmastrapas: @maximilianwerk check out the new tests with the fixtures

maximilianwerk: Much better! Thanks.

maximilianwerk (approved):
LGTM

PR #224, opened 2021-11-23
Feat evaluators

maximilianwerk:
First attempt on Evaluation with our dataset. Missing:

- conversion from class dataset.
- having summary with averaged NDCG etc.


finetuner/tuner/evaluation.py
```diff
@@ -0,0 +1,83 @@
...
+    from ..helper import AnyDNN
+
+
+class Evaluator:
+    def __init__(
+        self,
+        eval_data,
+        catalog,
+        embed_model: Optional[AnyDNN] = None,
+        override_original: bool = False,
```
tadejsv, line 19: What is the purpose of this argument?

finetuner/tuner/evaluation.py
```diff
@@ -0,0 +1,83 @@
...
+        self._eval_data = eval_data
+        self._catalog = catalog
+        self._override_original = override_original
+        self._summary_docs = self._parse_eval_docs()
+        self._limit = limit
+
+    def _parse_eval_docs(self):
+        to_be_scored_docs = DocumentArray()
+        for doc in self.eval_docs:
+            relevancies = [
```
tadejsv, line 42: Do we need this? All our training currently operates on the idea that documents are either similar or they are not - we have no such thing as a relevance score. Not sure we will ever need it

finetuner/tuner/evaluation.py
```diff
@@ -0,0 +1,83 @@
...
+            )
+            to_be_scored_docs.append(d)
+
+        return to_be_scored_docs
+
+    def _refresh_matches(self):
+        embed(self._eval_data, self._embed_model)
+
+        for doc in self._summary_docs:
+            doc.embedding = self.eval_docs[doc.id].embedding
```
tadejsv, line 66: Why not embed summary docs directly?

finetuner/tuner/evaluation.py
```diff
@@ -0,0 +1,83 @@
...
+
+        self._summary_docs.matches.clear()
+        self._summary_docs.match(self._catalog, limit=self._limit)
```
tadejsv, line 69: We just assume catalog already has embeddings?

finetuner/tuner/evaluation.py
maximilianwerk: We might have on an evaluation dataset. This doesn't mean you have to utilize it in training.

finetuner/tuner/evaluation.py
maximilianwerk: We thought about it. then we would need to copy the `blob` as well, which will double the memory footprint. when they are images or videos that might be non-neglectable, right?

finetuner/tuner/evaluation.py
maximilianwerk: Ouch, really good observation! No catalog also needs to be embedded every time.

finetuner/tuner/evaluation.py
maximilianwerk: We had plans for it and forgot to remove it.

finetuner/tuner/evaluation.py
tadejsv: I see. So I guess the whole need for copying eval docs to summary docs is because you will end up replacing matches, right?

maximilianwerk: Yes.

finetuner/tuner/evaluation.py
```diff
@@ -0,0 +1,83 @@
...
+from ..embedding import embed
+
+if TYPE_CHECKING:
+    from ..helper import AnyDNN
+
+
+class Evaluator:
+    def __init__(
+        self,
+        eval_data,
```
JoanFM, line 16: Some typehinting would be really useful

finetuner/tuner/evaluation.py
maximilianwerk: And we create the `relevancies` list only once and don't want to pollute the `eval_data` with it.

finetuner/tuner/metrics.py
```diff
@@ -0,0 +1,162 @@
+from typing import Optional, Dict, List, Tuple
```
hanxiao, line 1: this file is mostly redundant, as it is basically doing 3 lines of code:

```python
import jina.math.evaluations
rel = _doc_to_relevance(doc)[0]
getattr(jina.math.evaluation, metric_str)(rel, *args)
```

hanxiao (dismissed):
`finetuner/tuner/metrics.py` is redundant to me

finetuner/tuner/evaluation.py
gmastrapas: Added `eval_data: 'DocumentSequence'`, `catalog: 'DocumentSequence'`

finetuner/tuner/metrics.py
gmastrapas: @hanxiao are you ok with the current format? Or should I discard `metrics.py` and go with the following:

```
import jina.math.evaluations
rel, max_rel = _doc_to_relevance(doc)
if metric_str == 'recall' or metric_str == 'f1score':
    return getattr(jina.math.evaluation, metric_str)(rel, max_rel, *args)
return getattr(jina.math.evaluation, metric_str)(rel, *args)
```

finetuner/tuner/metrics.py
```diff
@@ -0,0 +1,164 @@
...
+    "r_precision": r_precision,
+    "precision": precision,
+    "recall": recall,
+    "f1score": f1score,
+    "average_precision": average_precision,
+    "hit": hit,
+    "reciprocal_rank": reciprocal_rank,
+    "dcg": dcg,
+    "ndcg": ndcg,
+}
```
bwanglzu, line 164: i have one suggestion, use `metrics` as the classmethod of `Evaluator`, such as 

```python
class Evaluator:

    @classmethod
    def list_available_metrics(cls):
        return ...
```

when user don't know what to use as evaluation function, this might provide some hints.

bwanglzu (commented):
one minor comment

finetuner/tuner/metrics.py
gmastrapas: good point, added!

finetuner/tuner/metrics.py
maximilianwerk: If these four lines fully functionally replace this file, let's replace it.

finetuner/tuner/evaluation.py
```diff
@@ -0,0 +1,132 @@
...
+        :param distance_metric: which distance metric to use when matching documents
+        :param limit: limit the number of results during matching.
+        :return: None.
+        """
+        if metrics:
+            self._metrics = {}
+            for metric in metrics:
+                if metric not in METRICS:
+                    raise ValueError(f"Unknown metric '{metric}'")
+                self._metrics[metric] = METRICS[metric]
```
maximilianwerk, line 40: We should either test functionality of just choosing specific metrics or remove the `metrics` argument completely. Thinking about it, I am in favor of removing it and rather add it back in the future.

gmastrapas: `metrics` argument removed from constructor

finetuner/tuner/metrics.py
gmastrapas: Removed wrapping logic

finetuner/tuner/evaluation.py
```diff
@@ -0,0 +1,160 @@
...
+        embed_model: Optional['AnyDNN'] = None,
+        distance_metric: str = 'cosine',
+        limit: int = 20,
+    ):
+        """
+        Build an Evaluator object that can be used to evaluate an embedding model on a retrieval task
+
+        :param eval_data: a sequence of documents. Each document should contain ground truth matches under
+            ``doc.matches`` and relevance scores under ``doc.tags[__default_tag_key__]['label']``
+        :param catalog: a sequence of documents, against whist the eval docs will be matched.
```
bwanglzu, line 49: what is this `whist`?

finetuner/tuner/evaluation.py
```diff
@@ -0,0 +1,160 @@
...
+        :param embed_model: the embedding model to use, in order to extract document representations.
+        :param distance_metric: which distance metric to use when matching documents
```
bwanglzu, line 51: can you document the allowed metrics here as well?

finetuner/tuner/evaluation.py
```diff
@@ -0,0 +1,160 @@
...
+        Convert the evaluation docs to the internal representation used by the Evaluator
+        """
+        to_be_scored_docs = DocumentArray()
+        for doc in self._eval_data:
+            relevancies = [
+                (m.id, m.tags[__default_tag_key__]['label'])
+                for m in doc.matches
+                if m.tags[__default_tag_key__]['label'] > 0
+            ]
+            sorted_relevancies = sorted(relevancies, key=lambda x: x[1])
```
bwanglzu, line 87: no need to change, but just to let you know, we allow something like:

```python

matches = doc.matches.sort(key=lambda d: d.tags['id'], reverse=True)
for d in filter(lambda d: d.tags[__default_tag_key__]['label'] > 0, doc.matches):
    print(d)
```

checkout the docs [here](https://docs.jina.ai/fundamentals/document/documentarray/?highlight=filter#filter)

bwanglzu (changes requested):
some comments

finetuner/tuner/evaluation.py
tadejsv: This changed recently, now `__default_tag_key__='finetuner_label'`, so I think no need for `'label'`.

finetuner/tuner/evaluation.py
gmastrapas: it's `which` 😄

finetuner/tuner/evaluation.py
gmastrapas: changed to `[__default_tag_key__]` instead of `[__default_tag_key__]['label']`

finetuner/tuner/evaluation.py
gmastrapas: done

finetuner/tuner/evaluation.py
```diff
@@ -0,0 +1,161 @@
...
+        catalog: 'DocumentSequence',
+        embed_model: Optional['AnyDNN'] = None,
+        distance: str = 'cosine',
+        limit: int = 20,
+    ):
+        """
+        Build an Evaluator object that can be used to evaluate an embedding model on a retrieval task
+
+        :param eval_data: A sequence of documents. Each document should contain ground truth matches under
+            ``doc.matches`` and relevance scores under ``doc.tags[__default_tag_key__]['label']``
```
bwanglzu, line 48: align changes in docstring as well

finetuner/tuner/evaluation.py
```diff
@@ -0,0 +1,161 @@
...
+        self._distance = distance
+        self._limit = limit
+
+    @staticmethod
+    def _doc_to_relevance(doc: Document) -> Tuple[List[int], int]:
+        """
+        Convert a Jina document to a relevance representation
+        :param doc: A Jina document. The matched document identifiers, predicted by the search system, are expected to
+            be under ``doc.matches``. The expected documents identifiers along with the relevance scores, given by the
+            user as matching ground truth, should be under ``doc.tags[finetuner.__default_tag_key__]['targets']``.
```
bwanglzu, line 69: if now label is `__finetuner_label__`, how does it affects this?

finetuner/tuner/evaluation.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        List available metrics
+        """
+        return list(METRICS.keys())
+
+    def evaluate(self, label: str = 'metrics') -> Dict[str, float]:
+        """
+        Run evaluation
+        :param label: Per document metrics are written in each evaluation document under
+            ``doc.tags[__default_tag_key__][label]``.
```
bwanglzu, line 139: apply changes to docstring as well

bwanglzu (changes requested):

finetuner/tuner/evaluation.py
gmastrapas: this actually holds. Metrics will be written in `doc.tags["finetuner_label"]["metrics"]` when `label = "metrics"`

finetuner/tuner/evaluation.py
gmastrapas: in the internal docs that the evaluator holds (`self._summary_docs`) ground truth matches will be stored in `doc.tags["finetuner_label"]["targets"]`

finetuner/tuner/evaluation.py
gmastrapas: yep, good catch

bwanglzu (approved):
LGTM!

finetuner/tuner/evaluation.py
gmastrapas: resolved

gmastrapas:
@hanxiao can you review the changes wrt the `metrics.py` file?

PR #230, opened 2021-11-25
feat(tailor): support freeze by layer names

bwanglzu:
in this PR, we added a `freeze_layers` parameter as argument of `to_embedding_layer`, which takes a list of `str`, allow user to freeze model by specific layer names. If user didn't specify `freeze_layers` and `freeze` is `True`, we freeze all layers. 

we also lifted restriction on output dimensionality, to allow any layer of `embedding_layers` serve as feature extractor.

Next step:
1. flatten embedding layer if last layer is > 2d.
2. always attach a new output layer as defined by `output_dim`.


tadejsv (approved):
LGTM

PR #231, opened 2021-11-25
feat: callbacks

tadejsv:
This introduces callbacks. Here are the main changes:

- Callbacks are introduced using the `BaseCallback` class. They are modeled based on the keras callbacks.
- Callbacks receive the tuner instance as argument - this is giving them a lot of flexibility, which we may need in the future. However, most of the time callbacks will interact with the environment using the `.state` attribute of the tuner, which is an instance of the new `TunerState` class. It contains basic info on the run (current epoch, batch idx, los, etc.)
- Progress bar is transformed into a callback. Actually, I have found that working with Jina's progress bar is very difficult, so I have implemented a custom rich progress bar. I find the new one to be a large improvement
- I got rid of the `Summary` class. This currently introduces a small issue, as there is no built-in way to show progress. This will be addressed immediately with introduction of Tensorboard (and other) callback.

TODO
- [x] Fix existing test
- [x] Add basic callback tests


finetuner/tuner/keras/__init__.py
```diff
@@ -173,24 +145,31 @@ def fit(
...
-                )
-                m_train_loss += lt
+
+                # Setting here as re-shuffling can change number of batches
+                self.state.epoch = epoch
+                self.state.num_batches_train = train_dl.get_size()
+
+                self._trigger_callbacks('on_epoch_begin')
+
+                self._trigger_callbacks('on_train_begin')
```
bwanglzu, line 160: ~i think `on_train_begin` should be moved out of the for loop, right? otherwise it's identical to `on_epoch_begin`~

now I understand, we have another `on_fit_begin`, then can we remove `on_train_begin` to something like `on_train_epoch_begin`?

bwanglzu (commented):
one thing i think we need to decide is the call back names.

`on_fit_begin`, `on_train_begin` might be confusing, i suggest we go with `on_train_begin`, `on_train_epoch_begin` and `on_train_batch_begin`, we stick with `train` instead of `fit`.

finetuner/tuner/keras/__init__.py
```diff
@@ -62,60 +61,41 @@ def _get_default_optimizer(self, learning_rate: float) -> Optimizer:
...
-            description,
-            message_on_done=lambda: f'{train_loss} | {_summary}',
-            total_length=data.get_size(),
-        ) as p:
-            for inputs, labels in data.get_dataset():
-                embeddings = self._embed_model(inputs)
-                loss = self._loss(embeddings, labels)
+        for idx, (inputs, labels) in enumerate(data.get_dataset()):
+            self.state.batch_index = idx
+            self._trigger_callbacks('on_val_batch_begin')
```
bwanglzu, line 90: if we didn't implement a certain method in a callback, it will go for a silent pass?

bwanglzu (commented):
minor comment

finetuner/tuner/keras/__init__.py
tadejsv: You mean to rename it? Because `on_train_begin` executes exactly at the start of each training epoch

finetuner/tuner/keras/__init__.py
tadejsv: Then it defaults to method of `BaseCallback` which is a silent pass

finetuner/tuner/keras/__init__.py
bwanglzu: yes, how do you think, i find the keras naming very clear: 
6619160E-3BD4-4F56-8DAD-C150696FED6E

tadejsv: Good point, I will rename it

bwanglzu (approved):
LGTM!

PR #235, opened 2021-11-29
feat implement hard mining 233

gvondulong:
This PR introduces hard mining for triplet learning.


tadejsv:
In general, you will need to revert all changes where you replace single quote by double quote - I suspect that happened because you did not use the `-S` options when running `black`

PR #236, opened 2021-11-29
feat: miners for hard triplets

gvondulong:
This PR introduces multiple hard triplet miners.


finetuner/tuner/pytorch/miner.py
```diff
@@ -1,6 +1,10 @@
+from cProfile import label
+from multiprocessing.sharedctypes import Value
```
tadejsv, line 2: Please clean up unneeded imports

finetuner/tuner/pytorch/miner.py
```diff
@@ -59,6 +63,99 @@ def mine(
...
+        :param distances: A tensor matrix of pairwise distance between each two item
+            embeddings
+
+        :return: three 1D tensors, holding the anchor index, positive index and
+            negative index of each triplet, respectively
+        """
+        # distances = torch.Tensor(
+        #     np.array([[0, 184, 222], [184, 0, 45], [222, 45, 0]])
+        # )  # distances[:3, :3]
+        # labels = labels[:3]
```
tadejsv, line 106: Please clean up comments

finetuner/tuner/pytorch/miner.py
```diff
@@ -59,6 +63,99 @@ def mine(
         diffs = matches ^ 1
 
         matches.fill_diagonal_(0)
+
+        triplets = matches.unsqueeze(2) * diffs.unsqueeze(1)
+
+        return torch.where(triplets)
+
+
+class TripletHardMiner(TripletMiner):
```
tadejsv: Can you add a docstring for the miner, describing what it does and what the availible strategies are

tests/unit/tuner/pytorch/test_miner.py
```diff
@@ -101,6 +102,44 @@ def test_triplet_miner(labels):
     np.testing.assert_equal(neg_ind.numpy(), true_neg_ind)
 
 
+def test_triplet_easy_hard_miner(labels):
```
tadejsv, line 105: We will need test for all strategies

finetuner/tuner/pytorch/miner.py
```diff
@@ -59,6 +63,99 @@ def mine(
         diffs = matches ^ 1
 
         matches.fill_diagonal_(0)
+
+        triplets = matches.unsqueeze(2) * diffs.unsqueeze(1)
+
+        return torch.where(triplets)
+
+
+class TripletEasyHardMiner(TripletMiner):
```
tadejsv, line 72: Please provide a docstring. Also, do you need to subclass `TripletMiner` - it doesn't seem to share any methods with it.

finetuner/tuner/pytorch/miner.py
```diff
@@ -59,6 +63,99 @@ def mine(
...
+    def __init__(self, strategy="hard"):
+        self.strategy = strategy
+
+    def _get_per_row_min(
+        self, dist_mat: torch.Tensor, semihard_tsh: torch.Tensor = None
```
tadejsv, line 77:
```suggestion
        self, dist_mat: torch.Tensor, semihard_tsh: Optional[torch.Tensor] = None
```

finetuner/tuner/pytorch/miner.py
```diff
@@ -59,6 +63,99 @@ def mine(
...
+    ):
+        non_zero_mask = dist_mat > 0
+
+        # Prevent wrong neg. samples from being extracted
+        row_max = torch.max(dist_mat, dim=1, keepdim=True)[0]
+        dist_mat += (row_max + 1) * torch.logical_not(non_zero_mask)
```
tadejsv, line 83: Why is adding row max needed?

finetuner/tuner/pytorch/miner.py
```diff
@@ -59,6 +63,99 @@ def mine(
...
+        :return: three 1D tensors, holding the anchor index, positive index and
+            negative index of each triplet, respectively
+        """
+        # distances = torch.Tensor(
+        #     np.array([[0, 184, 222], [184, 0, 45], [222, 45, 0]])
+        # )  # distances[:3, :3]
+        # labels = labels[:3]
+        np.random.seed(1234)
+        distances = torch.Tensor(np.round_(np.random.random((6, 6)), 1))
+        distances -= distances * torch.eye(len(distances))
```
tadejsv, line 109: This should be cleaned up

tadejsv (changes requested):

JoanFM (commented):
Just as a general question, how ks it decided that a case is hard or not?

gvondulong:
> 
In the image below, all negative samples that fall within the red area are hard create a hard triplet. If a negative sample falls within the orange are, it is considered semihard. Green follows the same logic. 

In the code: 
- For hard triplets, we take a negative sample, that is close, and for the same anchor we take all positive samples that are further away from the negative one. 
- For easy triplets, we basically do the same in reverse. 
- For semihard, we take a negative sample and then take one positive sample that is still closer to the anchor, than the negative sample we picked. 
image

bwanglzu:
> For hard triplets, we take a negative sample, that is close, and for the same anchor we take all positive samples that are further away from the negative one.

question : how do we define close? cosine similarity, or it is `margin` involved?

JoanFM (commented):
@tadejsv isnt strategy something that we can add to the miner via composition and not inheritance?

finetuner/tuner/pytorch/miner.py
gvondulong: Because I am getting the minimum per row. Since there are potentially zeros in the rows, I need to overwrite them with a larger value to make sure they are not extracted.

finetuner/tuner/pytorch/miner.py
```diff
@@ -59,11 +62,151 @@ def mine(
         diffs = matches ^ 1
 
         matches.fill_diagonal_(0)
+
         triplets = matches.unsqueeze(2) * diffs.unsqueeze(1)
 
         return torch.where(triplets)
 
 
+class BaseClassEasyHardMiner:
```
gvondulong, line 71: I think it would be flatter/more pythonic, if this was simply a function, called smth. along the lines of: `apply_easy_hard_strategy`?
Would you agree?

finetuner/tuner/pytorch/miner.py
```diff
@@ -59,11 +62,151 @@ def mine(
...
+        matches, diffs = self.apply_strategy(matches, diffs, distances)
+        triplets = matches.unsqueeze(2) * diffs.unsqueeze(1)
+
+        return torch.where(triplets)
+
+
+class SiameseEasyHardMiner(BaseClassMiner[torch.Tensor], BaseClassEasyHardMiner):
+    def __init__(self, strategy: str = 'hard'):
+        BaseClassEasyHardMiner.__init__(self, strategy)
+
```
gvondulong, line 171: Will add the docstrings, once we settled on the implementation.

finetuner/tuner/pytorch/miner.py
tadejsv: Yes, I would do this

gvondulong:
> > For hard triplets, we take a negative sample, that is close, and for the same anchor we take all positive samples that are further away from the negative one.
> 
> question : how do we define close? cosine similarity, or it is `margin` involved?

Distances are given inside the miner, but in principle any distance metric is possible.

hanxiao (changes requested):
why miner has to be so tightly coupled with torch API? this makes porting to Keras/Paddle unnecessarily hard. Can we simply use numpy API to do the same? I'm not seeing anywhere using GPU anyway, so there is no obvious reason to me to use Pytorch.

tadejsv:
@hanxiao GPU is used everywhere the tensors are

gvondulong:
> why miner has to be so tightly coupled with torch API? this makes porting to Keras/Paddle unnecessarily hard. Can we simply use numpy API to do the same? I'm not seeing anywhere using GPU anyway, so there is no obvious reason to me to use Pytorch.

It is true that the GPUs should be used everywhere, since the input tensors will be on GPU during training. 
But I agree that we should think about sacrificing a little performance (hard mining is not really heavy) for reusability. 
Having the implementation in numpy, would alleviate the trouble of implementing it three times.

hanxiao:
in general, i'd like to first see this strategy really works, meaning that it can considerably improve the finetuned performance.

After we are sure about the performance improvement, we can think of generalize it to Keras and Paddle

JoanFM (changes requested):
The structure of the code seems to miss the concept of `pytorch` as `metadata` language and the concept of `pytorch` as the underlying model.

This needs to be clarified and tested how it works for tuning on all frameworks.

JoanFM (changes requested):
`miner` should go under `tuner` and not any specific framework

JoanFM (commented):
This is not what was agreed right?

gvondulong:
> This is not what was agreed right?

Can you elaborate?

JoanFM:
> > This is not what was agreed right?
> 
> Can you elaborate?

We can merge now, but as soon as a new mining strategy is impleemnted, we move to only implement `mining` in `torch` as it is our meta-language but not because it is tight to train torch models

gvondulong:
> > > This is not what was agreed right?
> > 
> > 
> > Can you elaborate?
> 
> We can merge now, but as soon as a new mining strategy is impleemnted, we move to only implement `mining` in `torch` as it is our meta-language but not because it is tight to train torch models

Well, the intention behind the refactor, in the last commits, was to separate the mining form the tuners. Now the `StrategicMiningHelper` class sits in `tuner` and is used in each submodule across all frameworks. There is now a wrapper, written in each of the frameworks, encloses the mining procedure (written in torch). Hope this answers the question. :)

finetuner/tuner/pytorch/miner.py
```diff
@@ -59,6 +60,138 @@ def mine(
...
+            negative index of each triplet, respectively
+        """
+        assert len(distances) == len(labels)
+
+        labels1, labels2 = labels.unsqueeze(1), labels.unsqueeze(0)
+        matches = (labels1 == labels2).byte()
+        diffs = matches ^ 1
+
+        matches.fill_diagonal_(0)
+        print("TORCH MATCHES/DIFFS: ", matches, diffs)
```
tadejsv, line 189: This should be removed

finetuner/tuner/paddle/miner.py
```diff
@@ -97,6 +194,82 @@ def mine(
...
+        # Apply mining strategy
+        updated_matches, updated_diffs = self.strategic_mining_helper.apply_strategy(
+            torch.Tensor(matches.numpy()),
+            torch.Tensor(diffs.numpy()),
+            torch.Tensor(distances.numpy()),
+            to_numpy=True,
+        )
+
+        matches = paddle.Tensor(updated_matches)
+        diffs = paddle.Tensor(updated_diffs)
```
tadejsv, line 260: You are not properly setting device. Also, same about `Tensor`

finetuner/tuner/paddle/miner.py
```diff
@@ -97,6 +194,82 @@ def mine(
...
+        if labels.size == 0:
+            return _empty_tensor(), _empty_tensor(), _empty_tensor()
+
+        labels1, labels2 = labels.unsqueeze(1), labels.unsqueeze(0)
+        matches = paddle.cast(labels1 == labels2, 'int32')
+        diffs = 1 - matches
+
+        # Apply mining strategy
+        updated_matches, updated_diffs = self.strategic_mining_helper.apply_strategy(
+            torch.Tensor(matches.numpy()),
```
tadejsv, line 253: Using `Tensor` instead of `tensor` will convert everything to `float32` - but this is not the desired type, we want `uint8`. Also, you are not setting the device

finetuner/tuner/paddle/miner.py
```diff
@@ -63,6 +65,101 @@ def mine(
...
+
+        l1, l2 = labels.unsqueeze(1), labels.unsqueeze(0)
+        matches = paddle.cast(l1 == l2, 'int32')
+        diffs = 1 - matches
+        matches = paddle.triu(matches, 1)
+        diffs = paddle.triu(diffs)
+
+        # Apply mining strategy
+        updated_matches, updated_diffs = self.strategic_mining_helper.apply_strategy(
+            torch.Tensor(matches.numpy()),
```
tadejsv, line 127: Same here regarding conversion to tensor

finetuner/tuner/paddle/miner.py
```diff
@@ -63,6 +65,101 @@ def mine(
...
+            torch.Tensor(diffs.numpy()),
+            torch.Tensor(distances.numpy()),
+            to_numpy=True,
+        )
+
+        matches = paddle.Tensor(updated_matches)
```
tadejsv, line 133: Same here regarding device/type

finetuner/tuner/miner/mining_strategies.py
```diff
@@ -0,0 +1,232 @@
...
+            if self.pos_strategy != 'all':
+                match_mat, _ = self._update_pos_mat(
+                    match_mat, dist_mat, self.pos_strategy
+                )
+            if self.neg_strategy != 'all':
+                diff_mat, _ = self._update_neg_mat(
+                    diff_mat, dist_mat, self.neg_strategy
+                )
+        if to_numpy:
+            return match_mat.cpu().detach().numpy(), diff_mat.cpu().detach().numpy()
```
tadejsv, line 230: No point in `detach`

finetuner/tuner/pytorch/miner.py
```diff
@@ -59,6 +60,138 @@ def mine(
...
+        matches, diffs = self.strategic_mining_helper.apply_strategy(
+            matches, diffs, distances
+        )
+        print("TORCH UPDATED MATCHES/DIFFS: ", matches, diffs)
```
tadejsv, line 193: Same here

gvondulong:
@tadejsv on the issue you raised with converting the tensors to float: unfortunately I need to compare the values to `inf` tp make the logic work: So I fear that mining will need to work with float values.

finetuner/tuner/keras/miner.py
```diff
@@ -63,6 +144,82 @@ def mine(
...
+        # Apply mining strategy
+        updated_matches, updated_diffs = self.strategic_mining_helper.apply_strategy(
+            torch.Tensor(matches.numpy()),
+            torch.Tensor(diffs.numpy()),
+            torch.Tensor(distances.numpy()),
+            to_numpy=True,
+        )
+        # Send updated tensors to GPU if available
+        if len(tf.config.list_physical_devices('GPU')) > 0:
+            with tf.device('/CPU:0'):
```
tadejsv, line 208: We can remove this - TF device is set globally

finetuner/tuner/keras/miner.py
```diff
@@ -63,6 +144,82 @@ def mine(
...
+        matches = tf.cast(l1 == l2, tf.uint8)
+        diffs = matches ^ 1
+
+        matches = tf.linalg.set_diag(matches, diagonal=tf.zeros_like(labels, tf.uint8))
+
+        # Apply mining strategy
+        updated_matches, updated_diffs = self.strategic_mining_helper.apply_strategy(
+            torch.Tensor(matches.numpy()),
+            torch.Tensor(diffs.numpy()),
+            torch.Tensor(distances.numpy()),
```
tadejsv, line 203: Here you need to set device

tadejsv (commented):
Last two things, otherwise good to merge

tadejsv (approved):
LGTM

JoanFM (approved):

PR #237, opened 2021-11-29
feat: add wandb logger

tadejsv:
This adds support for logging with WandB - when we add learning rate scheduling (and proper evaluation) this will be expanded.

Here's how to use it

```python
import finetuner
from finetuner.tuner.callback import WandBLogger

finetuner.fit(
    embed_model,
    callbacks=[WandBLogger()],
    train_data=generate_fashion(num_total=2000),
    eval_data=generate_fashion(is_testset=True, num_total=1000),
)
```

Here's how the logs then look like in wandb

image

### DIFF
--- a/finetuner/tuner/callback/wandb_logger.py
+++ b/finetuner/tuner/callback/wandb_logger.py
@@ -57,7 +57,10 @@ class WandBLogger(BaseCallback):
         Called at the end of the evaluation batch.
         """
         avg_loss = np.mean(self._val_losses)
-        self.wandb_logger.log(data={'val_loss': avg_loss}, step=self._train_step)
+        self.wandb_logger.log(
+            data={'val_loss': avg_loss, 'epoch': tuner.state.epoch},
+            step=self._train_step,
+        )
         self._val_losses = []
 
     def on_fit_end(self, tuner: 'BaseTuner'):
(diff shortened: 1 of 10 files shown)


finetuner/tuner/callback/wandb_logger.py
```diff
@@ -0,0 +1,71 @@
...
+            step=self._train_step,
+        )
+        self._val_losses = []
+
+    def on_fit_end(self, tuner: 'BaseTuner'):
+        """
+        Called at the end of the ``fit`` method call, after finishing all the epochs.
+        """
+
+        self.wandb_logger.finish()
```
JoanFM, line 71: what does finish do? If it does not properly happen, and any exceptio happen, in what state does it end?

bwanglzu (approved):
Generally looks good to me, to me it's even better if we call it `WandBCallback` it's easier

finetuner/tuner/callback/wandb_logger.py
tadejsv: https://docs.wandb.ai/ref/python/finish

To tell wandb to finish uploading all data

JoanFM: then we are not properly passing the exit code or reporting failures

PR #244, opened 2021-11-29
fix(tuner): make copy of non writable blob

bwanglzu:


tadejsv:
I think this should be done in the base `Dataset` classes, probably by creating a default preprocessing function. Otherwise I would not implement this - it's just a warning

bwanglzu:
@tadejsv this only affects `pytorch`, i don't think it's suitable to add to base class. But I did clean up the code a bit, removed useless check. Now it should be a bit cleaner.

bwanglzu:
`setFlag(..)` will give an error now. With newer numpy version it's not doable since it's immutable.

tadejsv (approved):

PR #248, opened 2021-11-30
feat: lr scheduler

tadejsv:
Implements a learning rate scheduler

image

TODO

- [x] pytorch implementation
- [x] add to logging
- [x] keras implementation
- [x] paddle implementation
- [x] add tests

### DIFF
--- a/finetuner/__init__.py
+++ b/finetuner/__init__.py
@@ -34,7 +34,7 @@ def fit(
             ['AnyDNN'], Union['AnyOptimizer', Tuple['AnyOptimizer', 'AnyScheduler']]
         ]
     ] = None,
-    default_learning_rate: float = 1e-3,
+    learning_rate: float = 1e-3,
     scheduler_step: str = 'batch',
     device: str = 'cpu',
     preprocess_fn: Optional['PreprocFnType'] = None,
@@ -59,7 +59,7 @@ def fit(
             ['AnyDNN'], Union['AnyOptimizer', Tuple['AnyOptimizer', 'AnyScheduler']]
         ]
     ] = None,
-    default_learning_rate: float = 1e-3,
+    learning_rate: float = 1e-3,
     scheduler_step: str = 'batch',
     device: str = 'cpu',
     preprocess_fn: Optional['PreprocFnType'] = None,
@@ -90,7 +90,7 @@ def fit(
             ['AnyDNN'], Union['AnyOptimizer', Tuple['AnyOptimizer', 'AnyScheduler']]
         ]
     ] = None,
-    default_learning_rate: float = 1e-3,
+    learning_rate: float = 1e-3,
     scheduler_step: str = 'batch',
     device: str = 'cpu',
     preprocess_fn: Optional['PreprocFnType'] = None,
@@ -119,7 +119,7 @@ def fit(
             ['AnyDNN'], Union['AnyOptimizer', Tuple['AnyOptimizer', 'AnyScheduler']]
         ]
     ] = None,
-    default_learning_rate: float = 1e-3,
+    learning_rate: float = 1e-3,
     scheduler_step: str = 'batch',
     device: str = 'cpu',
     preprocess_fn: Optional['PreprocFnType'] = None,
--- a/finetuner/__init__.py
+++ b/finetuner/__init__.py
@@ -34,7 +34,7 @@ def fit(
             ['AnyDNN'], Union['AnyOptimizer', Tuple['AnyOptimizer', 'AnyScheduler']]
         ]
     ] = None,
-    learning_rate: float = 1e-3,
+    default_learning_rate: float = 1e-3,
     scheduler_step: str = 'batch',
     device: str = 'cpu',
     preprocess_fn: Optional['PreprocFnType'] = None,
@@ -59,7 +59,7 @@ def fit(
             ['AnyDNN'], Union['AnyOptimizer', Tuple['AnyOptimizer', 'AnyScheduler']]
         ]
     ] = None,
-    learning_rate: float = 1e-3,
+    default_learning_rate: float = 1e-3,
     scheduler_step: str = 'batch',
     device: str = 'cpu',
     preprocess_fn: Optional['PreprocFnType'] = None,
@@ -90,7 +90,7 @@ def fit(
             ['AnyDNN'], Union['AnyOptimizer', Tuple['AnyOptimizer', 'AnyScheduler']]
         ]
     ] = None,
-    learning_rate: float = 1e-3,
+    default_learning_rate: float = 1e-3,
     scheduler_step: str = 'batch',
     device: str = 'cpu',
     preprocess_fn: Optional['PreprocFnType'] = None,
@@ -119,7 +119,7 @@ def fit(
             ['AnyDNN'], Union['AnyOptimizer', Tuple['AnyOptimizer', 'AnyScheduler']]
         ]
     ] = None,
-    learning_rate: float = 1e-3,
+    default_learning_rate: float = 1e-3,
     scheduler_step: str = 'batch',
     device: str = 'cpu',
     preprocess_fn: Optional['PreprocFnType'] = None,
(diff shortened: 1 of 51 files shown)


finetuner/__init__.py
```diff
@@ -28,8 +29,13 @@ def fit(
     epochs: int = 10,
     batch_size: int = 256,
     loss: Union[str, 'AnyDNN'] = 'SiameseLoss',
-    optimizer: Optional['AnyOptimizer'] = None,
```
JoanFM, line 31: why u can't pass an optimizer object?

finetuner/tuner/base.py
```diff
@@ -42,13 +42,18 @@ def get_default_miner(self, is_session_dataset: bool) -> BaseMiner:
...
 
     def __init__(
         self,
         embed_model: Optional[AnyDNN] = None,
         loss: Union[BaseLoss, str] = 'SiameseLoss',
+        configure_optimizer: Optional[
+            Callable[[AnyDNN], Union[AnyOptimizer, Tuple[AnyOptimizer, AnyScheduler]]]
+        ] = None,
+        learning_rate: float = 1e-3,
+        scheduler_step: str = 'batch',
```
JoanFM, line 56: same for this

finetuner/tuner/pytorch/__init__.py
```diff
@@ -179,8 +186,16 @@ def fit(
         self.device = get_device(device)
         self._embed_model = self._embed_model.to(self.device)
 
-        # Create optimizer
-        self._optimizer = optimizer or self._get_default_optimizer(learning_rate)
+        # Create optimizer (and scheduler)
+        if self._configure_optimizer:
+            res = self._configure_optimizer(self._embed_model)
+            if isinstance(res, tuple):
+                self._optimizer = res[0]
```
JoanFM, line 193:
```suggestion
                self._optimizer, self._scheduler = res
```

finetuner/__init__.py
tadejsv: In python we move the model to a GPU if needed. However, if this is done after optimizer is initialized, this will break the optimizer

> If you need to move a model to GPU via .cuda(), please do so before constructing optimizers for it. Parameters of a model after .cuda() will be different objects with those before the call.
> In general, you should make sure that optimized parameters live in consistent locations when optimizers are constructed and used.

finetuner/tuner/base.py
```diff
@@ -42,13 +42,18 @@ def get_default_miner(self, is_session_dataset: bool) -> BaseMiner:
...
     state: TunerState
 
     def __init__(
         self,
         embed_model: Optional[AnyDNN] = None,
         loss: Union[BaseLoss, str] = 'SiameseLoss',
+        configure_optimizer: Optional[
+            Callable[[AnyDNN], Union[AnyOptimizer, Tuple[AnyOptimizer, AnyScheduler]]]
+        ] = None,
+        learning_rate: float = 1e-3,
```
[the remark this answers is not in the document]
tadejsv: So this is only the "default" learning rate, if the user does not provide an optimizer (in which case the learning rate stays constant). I think renaming this will suffice

finetuner/tuner/__init__.py
```diff
@@ -69,26 +75,34 @@ def fit(
         - ``TripletLoss`` for Triplet network
     :param num_items_per_class: Number of items from a single class to include in
         the batch. Only relevant for class datasets
+    :param configure_optimizer: A function that allows you to provide a custom
```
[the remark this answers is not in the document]
tadejsv: See above - basically, in pytorch to construct the optimizer you need to pass the parameters of the model

finetuner/tuner/base.py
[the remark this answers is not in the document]
JoanFM: The point is, `learning_rate` is an argument of `fit` not something that the `tuner` should be adapted to. Same tuner identity should be able to work with different `learning_rates`.

finetuner/__init__.py
JoanFM: The problem is that u are moving `fit` arguments to `attributes` of the `objects`. What a `Tuner` is should be determined by the model and the `loss` perhaps. `Optimizers` and `learning_rates` should be different call arguments possible with the same `object`.

JoanFM: Well no, the problem persists, and the factory idea is good, but they should remain call arguments and not attributes

tadejsv: I have thought about this, and I have decided to go against it. What is an "attribute" of Tuner and what is not is just semantics really - I can easily argue that because this is `Tuner` and not `Model` all things crucial to tuning need to go in `__init__`. 

I have instead a very concrete argument - all the things whose state can change during training should go into `__init__`. This is due to checkpointing - if my training is interrupted in epoch 5, and I want to later start from that position, I need a way of restoring that state. There are two ways of doing this:
1. Having the `fit` method already start with initialized states. In this case, I can, right after initializing the trainer, call `TrainingCheckpoint.load(tuner, path_to_checkpoint)`. This is my preferred option, as it decouples checkpoint from the trainer.
2. Having the checkpoint and training tightly coupled, so that you do `tuner.fit(load_checkpoint=path_to_checkpoint)`. I don't really like this option.

And the way this is done in different frameworks: 
- in keras you call `compile` on a model first (equivalent of our trainer init), passing optimizer with learning rate, metrics, etc. In `fit` you only pass data and related args
- in pytorch lightning you pass all the configuration (including things like number of epochs, device) in init, and pass model with data (optimizer  is specified as part of the model definition) and checkpoint path in `fit` - here they have checkpointing tightly coupled with the training method
- in pytorch ignite you have to create all objects before calling `Engine.run` - which takes only data and number of epochs.

Based on the prefered way of implementing checkpoints, and the observation that this is indeed how it is also done in other frameworks, I would go ahead with my implementation.

tests/integration/keras/test_lr.py
```diff
@@ -0,0 +1,40 @@
+import numpy as np
+import pytest
+import tensorflow as tf
+from finetuner.tuner.keras import KerasTuner
+
+
+@pytest.mark.parametrize('scheduler_step', ['batch'])
```
bwanglzu, line 7: guess you want to add `epoch` to here

finetuner/tuner/pytorch/__init__.py
```diff
@@ -197,6 +195,10 @@ def fit(
 
             self._trigger_callbacks('on_train_epoch_begin')
             self._train(train_dl)
+
+            if self._scheduler_step == 'epoch' and self._scheduler is not None:
```
bwanglzu, line 199:
```suggestion
            if self._scheduler and self._scheduler_step == 'epoch' :
```
if you want to check it exist in an if, should be checked first.

bwanglzu (changes requested):
i'm not sure about `configure_optimizer` the usage looks a bit wired to me. At least, the naming sounds wired.

One question I have is the `learning_rate` and `default_learning_rate`, does it means that if user provide a `learning_rate`, the `default..` will be overwritten? why change the name here?

tests/integration/keras/test_lr.py
tadejsv: I don't. Keras scheduler is stepped automatically - I mention this in docstrings

tadejsv:
@bwanglzu The naming is taken from [Pytorch lightning](https://pytorch-lightning.readthedocs.io/en/latest/common/lightning_module.html#configure-optimizers) where they do the same thing.

I see I forgot to rename `learning_rate` to `default_learning_rate` in some places, I will fix that

finetuner/tuner/pytorch/__init__.py
tadejsv: I don't see what improvement this brings?

bwanglzu: okay i thought you're checking scheduler, did not notice the second is `_scheduler_step`

bwanglzu (approved):
LGTM!

PR #249, opened 2021-11-30
feat: model checkpoint

### LINKED ISSUE #240: Model checkpoints as follow up from callback functions

azayz:
close #240 #259  #260


finetuner/tuner/callback/model_checkpoint.py
```diff
@@ -0,0 +1,134 @@
...
+        Called at the end of the training epoch.
+        """
+        if self.monitor == 'loss':
+            self._save_model(tuner)
+
+    def on_val_end(self, tuner: 'BaseTuner'):
+        """
+        Called at the end of the validation epoch.
+        """
+        if self.monitor == "val_loss":
```
JoanFM, line 90: single quote

finetuner/tuner/callback/model_checkpoint.py
```diff
@@ -0,0 +1,134 @@
...
+from finetuner.helper import get_framework
+from jina.logging.logger import JinaLogger
+
+if TYPE_CHECKING:
+    from ..base import BaseTuner
+
+
+class ModelCheckpointCallback(BaseCallback):
+    """
+    Callback to save model at some frequency (epochs or batchs)
```
JoanFM, line 14: this docstring does not seem accurate, `batches` do not seem to be considered at all.

JoanFM (changes requested):

tests/unit/tuner/callback/test_model_checkpoint.py
```diff
@@ -0,0 +1,177 @@
...
+            out_features=128,
+        ),
+        paddle.nn.ReLU(),
+        paddle.nn.Linear(in_features=128, out_features=32),
+    )
+    return embed_model
+
+
+def test_keras_model(keras_model: BaseTuner):
+    with tempfile.TemporaryDirectory() as tmpdirname:
```
maximilianwerk, line 55: You can use the `tmpdir` fixture. See for example [here](https://github.com/jina-ai/jina/blob/898d15731e2f1faf414b4641d9e438b0bef8f324/tests/unit/flow-construct/test_flow_visualization.py#L13)

tests/unit/tuner/callback/test_model_checkpoint.py
```diff
@@ -0,0 +1,177 @@
...
+            pytorch_model,
+            epochs=1,
+            train_data=generate_fashion(num_total=1000),
+            eval_data=generate_fashion(is_testset=True, num_total=200),
+            callbacks=[
+                ModelCheckpointCallback(filepath=tmpdirname, save_best_only=True)
+            ],
+        )
+
+        assert os.listdir(tmpdirname) == ['best_model']
```
maximilianwerk, line 113: All your tests use the `finetuner.fit` function. I'd prefer Unit tests to work on class level. Can you add at least one unittest at class level?

finetuner/tuner/callback/model_checkpoint.py
```diff
@@ -0,0 +1,134 @@
...
+
+        if mode not in ['auto', 'min', 'max']:
+            self.logger.warning(
+                'ModelCheckpoint mode %s is unknown, ' 'fallback to auto mode.', mode
+            )
+            mode = 'auto'
+
+        if mode == 'min':
+            self.monitor_op = np.less
+            self.best = np.Inf
```
maximilianwerk, line 67: All class variables should be private.

finetuner/tuner/callback/model_checkpoint.py
```diff
@@ -0,0 +1,134 @@
...
+        the maximization or the minimization of the monitored quantity.
+        For `val_acc`, this should be `max`, for `val_loss` this should be
+        `min`, etc. In `auto` mode, the mode is set to `max` if the quantities
+        monitored are 'acc' or start with 'fmeasure' and are set to `min` for
+        the rest of the quantities.
+        """
+        self.logger = JinaLogger(self.__class__.__name__)
+        self.filepath = filepath
+        self.save_best_only = save_best_only
+        self.monitor = monitor
```
maximilianwerk, line 53: All class variables should be private.

maximilianwerk (changes requested):
Some style improvements. The logic itself looks solid.

bwanglzu (changes requested):
i'm a bit rush, I'll continue review after finish some presentation

tests/unit/tuner/callback/test_model_checkpoint.py
```diff
@@ -0,0 +1,197 @@
...
+
+    checkpoint.on_val_end(tuner)
+
+    assert os.listdir(tmpdir) == ['saved_model_epoch_03']
+    assert set(os.listdir(os.path.join(tmpdir, 'saved_model_epoch_03'))) == {
+        'variables',
+        'assets',
+        'keras_metadata.pb',
+        'saved_model.pb',
+    }
```
bwanglzu, line 197: can you put different test into different folder? in keras/paddle/pytorch test folder

tests/unit/tuner/callback/test_model_checkpoint.py
```diff
@@ -0,0 +1,197 @@
...
+import paddle
+from finetuner.tuner.callback import ModelCheckpointCallback
+from finetuner.tuner.base import BaseTuner
+import finetuner
+from finetuner.toydata import generate_fashion
+import os
+import numpy as np
+from finetuner.tuner.pytorch import PytorchTuner
+from finetuner.tuner.keras import KerasTuner
+from finetuner.tuner.state import TunerState
```
bwanglzu, line 13: same problem

finetuner/tuner/callback/model_checkpoint.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        if self.__monitor == 'loss':
+            self._save_model(tuner)
+
+    def on_val_end(self, tuner: 'BaseTuner'):
+        """
+        Called at the end of the validation epoch.
+        """
+        if self.__monitor == 'val_loss':
+            self._save_model(tuner)
```
bwanglzu, line 98: I suggest we do 2 changes:

1. save only happens at the end of `epoch end`.
2. if we have `val_loss`, it's only make sense to save model which has a better validation loss, not training loss. otherwise, save best model with loss (incase user didn't specify val_data).
3. maybe we need to checkpoint more things, see how pytorch doing (here)[https://github.com/pytorch/examples/blob/537f6971872b839b36983ff40dafe688276fe6c3/imagenet/main.py#L237]

bwanglzu (changes requested):
left some comments

finetuner/tuner/callback/model_checkpoint.py
azayz: I think we need both epoch end and val end because at everytime we send current loss within the tuner state, if it's epoch end I think only training loss is sent and if its val end it's the validation loss that's sent

finetuner/tuner/callback/model_checkpoint.py
```diff
@@ -0,0 +1,141 @@
...
+from jina.logging.logger import JinaLogger
+from finetuner.helper import get_framework
+from .base import BaseCallback
+
+
+if TYPE_CHECKING:
+    from ..base import BaseTuner
+
+
+class ModelCheckpointCallback(BaseCallback):
```
tadejsv, line 13: I would remove the `Callback` from name

finetuner/tuner/callback/model_checkpoint.py
```diff
@@ -0,0 +1,141 @@
...
+        the maximization or the minimization of the monitored quantity.
+        For `val_acc`, this should be `max`, for `val_loss` this should be
+        `min`, etc. In `auto` mode, the mode is set to `max` if the quantities
+        monitored are 'acc' or start with 'fmeasure' and are set to `min` for
+        the rest of the quantities.
+        """
+        self.__logger = JinaLogger(self.__class__.__name__)
+        self.__filepath = filepath
+        self.__save_best_only = save_best_only
+        self.__monitor = monitor
```
tadejsv, line 54: One leading underscore is enough

finetuner/tuner/callback/model_checkpoint.py
```diff
@@ -0,0 +1,141 @@
...
+        :param monitor: if `monitor='loss'` best bodel saved will be according
+        to the training loss, if `monitor='val_loss'` best model saved will be
+        according to the validation loss
+        :param mode: one of {'auto', 'min', 'max'}. If `save_best_only=True`, the
+        decision to overwrite the current save file is made based on either
+        the maximization or the minimization of the monitored quantity.
+        For `val_acc`, this should be `max`, for `val_loss` this should be
+        `min`, etc. In `auto` mode, the mode is set to `max` if the quantities
+        monitored are 'acc' or start with 'fmeasure' and are set to `min` for
+        the rest of the quantities.
```
tadejsv, line 49: Please indent the descriptions of parameters, so that it is visually clear what parameter they belong to

finetuner/tuner/callback/model_checkpoint.py
```diff
@@ -0,0 +1,141 @@
...
+    def on_val_end(self, tuner: 'BaseTuner'):
+        """
+        Called at the end of the validation epoch.
+        """
+        if self.__monitor == 'val_loss':
+            self._save_model(tuner)
+
+    def _save_model(self, tuner):
+        if self.__save_best_only:
+            current = tuner.state.current_loss
```
tadejsv, line 102: One thing: the loss is only given as the loss per batch - but if you want to use `val_loss` what you need to do is to compute the average of it. To do this you'll need to hook into `on_val_batch_end` and collect the losses there

finetuner/tuner/callback/model_checkpoint.py
tadejsv: I think it should be on epoch end. The value in `current_loss` is simply the last loss computed - if there is validation, it will be loss of last validation loss. But as I explain in another comment, you actually need to collect validation losses for all batches and average them

tests/unit/tuner/callback/test_model_checkpoint.py
tadejsv: This is not resolved - please order the imports

finetuner/tuner/callback/model_checkpoint.py
```diff
@@ -0,0 +1,141 @@
...
+    - The frequency it should save at. Currently, the callback supports saving at
+        the end of every epoch, or after a fixed number of training batches.
+    """
+
+    def __init__(
+        self,
+        filepath: str = None,
+        save_best_only: Optional[bool] = False,
+        monitor: Optional[str] = 'val_loss',
+        mode: Optional[str] = 'auto',
```
tadejsv, line 34: Instead of `auto`, I would just use `None` as default - you have typed the argument as `Optional` anyway. So, what I mean is that `None` would replace `'auto'`

finetuner/tuner/callback/model_checkpoint.py
```diff
@@ -0,0 +1,141 @@
...
+                self.__best = -np.Inf
+            else:
+                self.__monitor_op = np.less
+                self.__best = np.Inf
+
+    def get_monitor_op(self):
+        return self.__monitor_op
+
+    def get_best(self):
+        return self.__best
```
tadejsv, line 84: These are unnecessary

finetuner/tuner/callback/model_checkpoint.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def __init__(
+        self,
+        filepath: str = None,
+        save_best_only: Optional[bool] = False,
+        monitor: Optional[str] = 'val_loss',
+        mode: Optional[str] = 'auto',
+    ):
+        """
+        :param filepath: string or `PathLike`, path to save the model file.
```
tadejsv, line 37: Can you rename this to something like `save_dir`, because it is the directory where to save. Naming it `filepath` can confuse people into thinking this is the (full) file name of saved model.

finetuner/tuner/callback/model_checkpoint.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def _get_file_path(self, tuner):
+        """
+        Returns the file path for checkpoint.
+        """
+
+        if self.__save_best_only:
+            file_path = os.path.join(
+                self.__filepath,
+                'best_model',
```
tadejsv, line 134: So what if I add multiple callbacks, watching different metrics? Then these checkpoints will overwrite each other?

tadejsv (changes requested):
One general problem: when saving the latest checkpoint, our intention is that we can resume training from that state if it is interrupted for some reason.

However, for that you would need to save the optimizer state as well (as a minimum). In Pytorch and I guess paddle this is easy, not sure about Keras.

finetuner/tuner/callback/model_checkpoint.py
azayz: Used in tests to check if mode functions correctly

tadejsv: You can access the private attributes directly in test, nothing wrong with that I would say

azayz: okay I'll remove them

finetuner/tuner/callback/model_checkpoint.py
azayz: So if I understand correctly the loss received in the `tuner state` in `on_epoch_end` and `on_val_end` is the loss of the last batch not the whole data ?

finetuner/tuner/callback/model_checkpoint.py
azayz: Okay I'll make changes accordingly

finetuner/tuner/callback/model_checkpoint.py
```diff
@@ -0,0 +1,160 @@
...
+        maximized or minimized.
+    - The frequency it should save at. Currently, the callback supports saving at
+        the end of every epoch, or after a fixed number of training batches.
+    """
+
+    def __init__(
+        self,
+        save_dir: str = None,
+        save_best_only: Optional[bool] = False,
+        monitor: Optional[str] = 'val_loss',
```
bwanglzu, line 35: when use `Optional`, why you set a default value? should be either `Optional[str] = None` or `str = 'val_loss'`

tests/unit/tuner/callback/test_model_checkpoint.py
```diff
@@ -0,0 +1,90 @@
...
+    assert checkpoint._best == np.Inf
+
+
+def test_mode_auto_max(tmpdir):
+
+    checkpoint = ModelCheckpoint(
+        save_dir=tmpdir, save_best_only=True, mode='auto', monitor='acc'
+    )
+    assert checkpoint._monitor_op == np.greater
+    assert checkpoint._best == -np.Inf
```
bwanglzu, line 67: can be parameterized

bwanglzu (changes requested):
some other comments

finetuner/tuner/callback/model_checkpoint.py
JoanFM: Yes, indeed does not seem to be respected

bwanglzu (approved):
LGTM!

finetuner/tuner/callback/best_model_checkpoint.py
```diff
@@ -0,0 +1,140 @@
...
+    An option this callback provides include:
+    - Definition of 'best'; which quantity to monitor and whether it should be
+        maximized or minimized.
+    """
+
+    def __init__(
+        self,
+        save_dir: str = None,
+        monitor: str = 'val_loss',
+        mode: Optional[str] = None,
```
tadejsv, line 30: Can we have this `auto` as default, and not `Optional`? Better explicit than implicit

finetuner/tuner/callback/training_model_checkpoint.py
```diff
@@ -0,0 +1,137 @@
...
+
+from finetuner.helper import get_framework
+from .base import BaseCallback
+
+
+if TYPE_CHECKING:
+    from ..base import BaseTuner
+
+
+class TrainingModelCheckpoint(BaseCallback):
```
tadejsv, line 15: We should rename to just `TrainingModelCheckpoint`, as it saves the whole training state, not just the model

finetuner/tuner/callback/training_model_checkpoint.py
```diff
@@ -0,0 +1,137 @@
...
+    An option this callback provides include:
+    - Definition of 'best'; which quantity to monitor and whether it should be
+        maximized or minimized.
+    """
+
+    def __init__(
+        self,
+        save_dir: str = None,
+        monitor: str = 'val_loss',
+        mode: Optional[str] = None,
```
tadejsv, line 30: I think these two arguments are not needed for this checkpoint

finetuner/tuner/paddle/__init__.py
```diff
@@ -221,7 +221,24 @@ def save(self, *args, **kwargs):
         :param args: Arguments to pass to ``paddle.save`` function
         :param kwargs: Keyword arguments to pass to ``paddle.save`` function
         """
-        paddle.save(self.embed_model.state_dict(), *args, **kwargs)
+        if hasattr(self, '_optimizer'):
```
tadejsv, line 224: In what case would it not have an optimizer?

finetuner/tuner/callback/best_model_checkpoint.py
```diff
@@ -0,0 +1,140 @@
...
+        Called at the end of the training epoch.
+        """
+        if self._monitor == 'loss':
+            self._save_model(tuner)
+            self._train_losses = []
+
+    def on_train_batch_end(self, tuner: 'BaseTuner'):
+        self._train_losses.append(tuner.state.train_loss)
+
+    def on_val_end(self, tuner: 'BaseTuner'):
```
tadejsv, line 86: I think we only need to save in `on_epoch_end`

finetuner/tuner/callback/training_model_checkpoint.py
azayz: you mean TrainingCheckpoint ?

finetuner/tuner/callback/best_model_checkpoint.py
azayz: what if we monitor validation loss and I want the best model after that epoch's validation

finetuner/tuner/callback/training_model_checkpoint.py
azayz: I agree

finetuner/tuner/callback/best_model_checkpoint.py
tadejsv: `on_epoch_end` occurs after `on_val_end`, so there you will have both training and validation loss available for comparison

finetuner/tuner/callback/best_model_checkpoint.py
```diff
@@ -0,0 +1,140 @@
...
+            mode = 'auto'
+
+        if mode == 'min':
+            self._monitor_op = np.less
+            self._best = np.Inf
+        elif mode == 'max':
+            self._monitor_op = np.greater
+            self._best = -np.Inf
+        else:
+            if 'acc' in self._monitor or self._monitor.startswith('fmeasure'):
```
tadejsv, line 68: Why `startswith` ?

finetuner/tuner/callback/training_model_checkpoint.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def on_epoch_end(self, tuner: 'BaseTuner'):
+        """
+        Called at the end of the training epoch.
+        """
+        self._save_model(tuner)
+        self._train_losses = []
+
+    def on_train_batch_end(self, tuner: 'BaseTuner'):
+        self._train_losses.append(tuner.state.train_loss)
```
tadejsv, line 49: This is not needed for this checkpoint

finetuner/tuner/callback/training_model_checkpoint.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def _save_model(self, tuner):
+        current = np.mean(self._train_losses)
+        if current is None:
+            self._logger.warning(
+                'Can save model only with train loss available, ' 'skipping.'
+            )
```
tadejsv, line 56: This is not needed for this checkpoint

finetuner/tuner/paddle/__init__.py
tadejsv: Okay, I think after my PR gets merged you could test normally and we can remove this

finetuner/tuner/callback/training_model_checkpoint.py
```diff
@@ -0,0 +1,81 @@
+import os
```
tadejsv, line 1: Please rename this file to `training_checkpoint.py`

finetuner/tuner/keras/__init__.py
```diff
@@ -196,8 +201,24 @@ def save(self, *args, **kwargs):
...
+        ) as f:
+            pickle.dump(state, f)
+
+    def load(self, fp, *args, **kwargs):
+        """Loads the embedding model, optimizer and updates the state epoch."""
+
+        self._embed_model = keras.models.load_model(fp, *args, **kwargs)
+        with open(os.path.join(fp, 'saved_state.pkl'), 'rb') as f:
+            loaded_state = pickle.load(f)
+        self.state.epoch = loaded_state['epoch']
```
tadejsv, line 221: Why do we need to change this? `save` is meant to be for model only, but here we are saving the whole trainer state?

finetuner/tuner/paddle/__init__.py
```diff
@@ -219,7 +221,24 @@ def save(self, *args, **kwargs):
...
+
+        paddle.save(state, *args, **kwargs)
+
+    def load(self, fp):
+        """Loads the embedding model, optimizer and updates the state epoch."""
+
+        checkpoint = paddle.load(fp)
+        self._embed_model.set_state_dict(checkpoint['state_dict'])
+        self._optimizer.set_state_dict(checkpoint['optimizer'])
+        self.state.epoch = checkpoint['epoch']
```
tadejsv, line 241: Same here - leave loading/saving of trainer state to the checkpoint, do not couple it with the trainer itself

finetuner/tuner/pytorch/__init__.py
```diff
@@ -222,7 +224,24 @@ def save(self, *args, **kwargs):
...
+
+        torch.save(state, *args, **kwargs)
+
+    def load(self, fp):
+        """Loads the embedding model, optimizer and updates the state epoch."""
+
+        checkpoint = torch.load(fp)
+        self._embed_model.load_state_dict(checkpoint['state_dict'])
+        self._optimizer.load_state_dict(checkpoint['optimizer'])
+        self.state.epoch = checkpoint['epoch']
```
tadejsv, line 244: Same here - leave loading/saving of trainer state to the checkpoint, do not couple it with the trainer itself

tadejsv (changes requested):

tadejsv (changes requested):
Good job with the checkpoints - I left some comments on that in the review, but besides that I think this is complete.

I would make the testing more rigorous, however. In unit tests, you never need to call `tuner.fit`, just manipulate the `tuner.state` and call the methods of the callback directly. 

For best model checkpoint, you need to test that when a new best score is acheived, the saving happens, and also that when a new best score is not acheived, the saving does not happen. This you need to test for both `min` and `max` mode.

For training checkpoint it is enough to simulate two epochs, and to make sure that saving happens each epoch. 

Btw, currently we are saving each epoch - but not deleting anything? This will completely clog up the disk if you train for say 100 epochs. This goes for best model checkpoint as well. It probably means you need to add some "tracking" of files that the checkpoint has saved. And then you need some argument on how (max) many checkpoints should be kept on disk

finetuner/embedding.py
```diff
@@ -90,6 +89,8 @@ def _set_embeddings_torch(
             else:
                 contents = b.contents
             batch_inputs = collate_fn(contents).to(device)
+            if isinstance(batch_inputs, torch.Tensor):
+                batch_inputs = batch_inputs.float()
```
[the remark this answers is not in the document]
azayz: yes this was quick easy fix, so far I don't know why the labeler test breaks, I'll need to investigate more

finetuner/tuner/evaluation.py
```diff
@@ -123,6 +122,7 @@ def _embed(self, docs: 'DocumentSequence'):
             preprocess_fn=self._preprocess_fn,
             collate_fn=self._collate_fn,
         )
+        docs.embeddings = docs.embeddings.astype('float64')
```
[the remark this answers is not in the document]
azayz: also this was to bypass the test_embedding, it requires embeddings to be of type float64 but in the set embedding the data type is float32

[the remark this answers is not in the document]
tadejsv: Okay, if something in evaluator requires use of `float64`, then this is good. But what part requires it?

finetuner/embedding.py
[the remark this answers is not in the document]
tadejsv: Got it yeah, I found out about the PR this morning

tadejsv (commented):
Some comments on tests - seems good otherwise, except for the issue with number of saved checkpoints when using TrainingCheckpoint

finetuner/tuner/callback/best_model_checkpoint.py
```diff
@@ -0,0 +1,146 @@
...
+    using `finetuner.fit()`
+
+    An option this callback provides include:
+    - Definition of 'best'; which quantity to monitor and whether it should be
+        maximized or minimized.
+    """
+
+    def __init__(
+        self,
+        save_dir: str = None,
```
JoanFM, line 31: proper type hinting

finetuner/tuner/callback/best_model_checkpoint.py
```diff
@@ -0,0 +1,146 @@
...
+        Saves the model weights depending on its framework.
+        """
+        if get_framework(tuner.embed_model) == 'keras':
+            tuner.save(
+                filepath=self._get_file_path(),
+            )
+        elif get_framework(tuner.embed_model) == 'torch':
+            tuner.save(f=self._get_file_path())
+        elif get_framework(tuner.embed_model) == 'paddle':
+            tuner.save(path=self._get_file_path())
```
JoanFM, line 123: why do u need to check the `framework` if the code is trhe same?

JoanFM (changes requested):

finetuner/tuner/callback/best_model_checkpoint.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+
+    def __init__(
+        self,
+        save_dir: str,
+        monitor: str = 'val_loss',
+        mode: str = 'auto',
+    ):
+        """
+        :param save_dir: string or `PathLike`, path to save the model file.
```
tadejsv, line 36: The description of the type does not match the type hint

finetuner/tuner/callback/best_model_checkpoint.py
```diff
@@ -0,0 +1,135 @@
...
+                'Can save best model only with %s available, ' 'skipping.',
+                self._monitor,
+            )
+        else:
+            if self._monitor_op(current, self._best):
+                self._best = current
+                tuner.save(
+                    self._get_file_path(),
+                )
+                self._logger.logger.info(
```
tadejsv, line 108: `self._logger.logger`? Shouldn't it be just `self._logger`?

finetuner/tuner/callback/training_checkpoint.py
```diff
@@ -0,0 +1,142 @@
...
+class TrainingCheckpoint(BaseCallback):
+    """
+    Callback to save model at every epoch or the last K epochs
+    `TrainingModelCheckpoint` is used in conjunction with training
+    using `finetuner.fit()`
+    """
+
+    def __init__(self, save_dir: str, last_k_epochs: int = None):
+        """
+        :param save_dir: string or `PathLike`, path to save the model file.
```
tadejsv, line 27: Same here

finetuner/tuner/callback/training_checkpoint.py
```diff
@@ -0,0 +1,142 @@
...
+
+if TYPE_CHECKING:
+    from ..base import BaseTuner
+
+
+class TrainingCheckpoint(BaseCallback):
+    """
+    Callback to save model at every epoch or the last K epochs
+    `TrainingModelCheckpoint` is used in conjunction with training
+    using `finetuner.fit()`
```
tadejsv, line 22: No need for "in conjunction with" - all callbacks are used like that

finetuner/tuner/callback/best_model_checkpoint.py
```diff
@@ -0,0 +1,135 @@
...
+
+if TYPE_CHECKING:
+    from ..base import BaseTuner
+
+
+class BestModelCheckpoint(BaseCallback):
+    """
+    Callback to save model at every epoch or the best model across all epochs
+    `BestModelCheckpoint` is used in conjunction with training
+    using `finetuner.fit()`
```
tadejsv, line 22: This description is not corrent anymore.

finetuner/tuner/callback/training_checkpoint.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+
+    def __init__(self, save_dir: str, last_k_epochs: int = None):
+        """
+        :param save_dir: string or `PathLike`, path to save the model file.
+        :param last_k_epochs: this parameter is an integer. It allows you yo save
+            only at the last k epochs and not on every epoch.
+            For example if `last_k_epochs = 3` and the total number of epochs is 10
+            we'll save only on epochs 8,9 and 10.
```
tadejsv, line 31: This is actually not how this feature should be implemented.

The point of this callback is to save the intermediate progress of training, in case something happens and the training is interrupted. There is no use case for saving only last 3 epochs.

What you want instead, is to save the _most recent_ 3 epochs. For example:
```
epoch 5 -> saved 5, 4, 3
epoch 6 -> saved 6, 5, 4 (here saved epoch 3 is deleted)
epoch 7 -> saved 7, 6, 5 (here saved epoch 4 is deleted)
```
And by default I would only keep the latest epoch (which is what most users would want)

tadejsv (changes requested):
The implementation of the number of epochs to keep in training checkpoint has to be adjusted - will do a full review of tests as well after that is done

finetuner/tuner/callback/training_checkpoint.py
azayz: Now I get it

finetuner/tuner/callback/best_model_checkpoint.py
```diff
@@ -0,0 +1,133 @@
...
+        if current is None:
+            self._logger.logger.warning(
+                'Can save best model only with %s available, ' 'skipping.',
+                self._monitor,
+            )
+        else:
+            if self._monitor_op(current, self._best):
+                self._best = current
+                tuner.save(
+                    self._get_file_path(),
```
tadejsv, line 104: Make this more compact
```suggestion
                    self._get_file_path()
```

finetuner/tuner/callback/best_model_checkpoint.py
```diff
@@ -0,0 +1,133 @@
...
+                self._logger.logger.info(f'Model didnt improve.')
+
+    def _get_file_path(self):
+        """
+        Returns the file path for checkpoint.
+        """
+
+        file_path = os.path.join(
+            self._save_dir,
+            'best_model_{}'.format(self._monitor),
```
tadejsv, line 119: Use `f`-string to make this compact, I think then it might all fit in one line
```suggestion
            f'best_model_{self._monitor}'
```

finetuner/tuner/callback/training_checkpoint.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def __init__(self, save_dir: str, last_k_epochs: int = None):
+        """
+        :param save_dir: string, path to save the model file.
+        :param last_k_epochs: this parameter is an integer. It allows you to save
+            on only the most recent epochs and not on every epoch.
+            For example if `last_k_epochs = 3` and the total number of epochs is 10
+            epoch 5 -> saved 5, 4, 3
+            epoch 6 -> saved 6, 5, 4 (here saved epoch 3 is deleted)
+            epoch 7 -> saved 7, 6, 5 (here saved epoch 4 is deleted)
```
tadejsv, line 32: I think this is not needed. Just say that as training progresses, only the most recenkt `k` checkpoints will be kept, and older ones removed each epoch

finetuner/tuner/callback/training_checkpoint.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def on_epoch_end(self, tuner: 'BaseTuner'):
+        """
+        Called at the end of the training epoch.
+        """
+        self._save_model_framework(tuner)
+        self._logger.logger.info(
+            f'Model trained for {tuner.state.epoch+1} epochs is saved!'
+        )
+        if self._last_k_epochs:
```
tadejsv, line 50: I think this part is not correctly implemented, and in any case way more lines than needed. Here's how I think this (deleting of excess checkpoints) could be done:

- create a property `saved_checkpoints`, which is a list with paths to saved checkpoints
- when creating a new checkpoints, append to the list
- when checking for deletion: check if `len(saved_checkpoints) > last_k_epochs` - if so, take the first element from `saved_checkpoints`, and delete the file, and after that remove the path from ``saved_checkpoints`

finetuner/tuner/callback/training_checkpoint.py
```diff
@@ -0,0 +1,143 @@
...
+if TYPE_CHECKING:
+    from ..base import BaseTuner
+
+
+class TrainingCheckpoint(BaseCallback):
+    """
+    Callback to save model at every epoch or the last K epochs
+    """
+
+    def __init__(self, save_dir: str, last_k_epochs: int = None):
```
tadejsv, line 24: Can you make the `last_k_epochs` parameter an integer with a default of 1?

tests/unit/tuner/callback/test_model_checkpoint.py
```diff
@@ -0,0 +1,102 @@
...
+
+@pytest.fixture(scope='module')
+def keras_model() -> BaseTuner:
+    embed_model = tf.keras.Sequential(
+        [
+            tf.keras.layers.Flatten(input_shape=(28, 28)),
+            tf.keras.layers.Dense(10, activation='relu'),
+        ]
+    )
+    return embed_model
```
tadejsv, line 30: Use only pytorch here, since the functionality you are testing does not depend on the framework

tests/unit/tuner/keras/test_model_checkpoint.py
```diff
@@ -0,0 +1,145 @@
...
+    after_stop_tuner.state = TunerState(epoch=0, batch_index=2, train_loss=1.1)
+
+    checkpoint = TrainingCheckpoint(save_dir=tmpdir)
+    checkpoint.on_epoch_end(before_stop_tuner)
+
+    checkpoint.load_model(
+        after_stop_tuner,
+        os.path.join(
+            tmpdir,
+            'saved_model_epoch_11',
```
tadejsv, line 85: Can you make this more compact

tests/unit/tuner/pytorch/test_model_checkpoint.py
```diff
@@ -0,0 +1,155 @@
...
+
+from finetuner.tuner.base import BaseTuner
+from finetuner.tuner.callback import BestModelCheckpoint, TrainingCheckpoint
+from finetuner.tuner.pytorch import PytorchTuner
+from finetuner.tuner.state import TunerState
+
+
+@pytest.fixture(scope='module')
+def pytorch_model() -> BaseTuner:
+    embed_model = torch.nn.Sequential(
```
tadejsv, line 17: Try to make this embedding model a one(or two)-liner. Same for other frameworks

tadejsv (commented):
Overall quite good, some final commnets

tests/unit/tuner/callback/test_model_checkpoint.py
azayz: I specifically used both because keras saves in a directory and pytorch saves in a file. So I use two ways to delete os.remove and shutil.rmtree I created two tests for pytorch and keras to be able to test both deletions

tadejsv: Got it

finetuner/tuner/callback/training_checkpoint.py
```diff
@@ -0,0 +1,143 @@
...
+            recent k checkpoints will be kept. Older checkpoints are deleted
+        """
+        self._logger = JinaLogger(self.__class__.__name__)
+        self._save_dir = save_dir
+        self._last_k_epochs = last_k_epochs
+        self._saved_checkpoints = []
+        if not save_dir:
+            raise ValueError(
+                '``save_dir`` parameter is mandatory. Pass it in parameters'
+            )
```
tadejsv, line 37: I think you can delete this - you already made `save_dir` a required parameter by not providing a default

finetuner/tuner/callback/best_model_checkpoint.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+        self._logger = JinaLogger(self.__class__.__name__)
+        self._save_dir = save_dir
+        self._monitor = monitor
+        self._train_losses = []
+        self._valid_losses = []
+        if not save_dir:
+            raise ValueError(
+                '``save_dir`` parameter is mandatory. Pass it in parameters'
+            )
```
tadejsv, line 54: I think you can delete this. You already made `save_dir` a required parameter by nor providing default

finetuner/tuner/callback/training_checkpoint.py
```diff
@@ -0,0 +1,143 @@
...
+        Saves the model weights, optimizer, scheduler and epoch
+        depending on its framework.
+        """
+        if get_framework(tuner.embed_model) == 'keras':
+            tuner.save(
+                filepath=self._get_file_path(tuner),
+            )
+            state = {
+                'epoch': tuner.state.epoch + 1,
+            }
```
tadejsv, line 69: Can you make this more compact

finetuner/tuner/callback/training_checkpoint.py
```diff
@@ -0,0 +1,143 @@
...
+                    'optimizer': tuner._optimizer.state_dict(),
+                    'scheduler': tuner._scheduler.state_dict(),
+                }
+            else:
+                state = {
+                    'epoch': tuner.state.epoch + 1,
+                    'state_dict': tuner.embed_model.state_dict(),
+                    'optimizer': tuner._optimizer.state_dict(),
+                    'scheduler': None,
+                }
```
tadejsv, line 88: I would refactor this to create a dict with epoch, state dict and optimizer first, and then add the scheduler to it only if it exists - will be clearer and less code

finetuner/tuner/callback/training_checkpoint.py
```diff
@@ -0,0 +1,143 @@
...
```
tadejsv, line 105: Same here

finetuner/tuner/state.py
```diff
@@ -16,3 +16,5 @@ class TunerState:
     current_loss: float = 0.0
```
tadejsv, line 16: If you add `train_loss` and `val_loss`, then delete `current_loss`

tests/integration/test_model_checkpoint.py
```diff
@@ -0,0 +1,52 @@
...
+    embed_model = torch.nn.Sequential(
+        torch.nn.Flatten(),
+        torch.nn.Linear(
+            in_features=28 * 28,
+            out_features=128,
+        ),
+        torch.nn.ReLU(),
+        torch.nn.Linear(in_features=128, out_features=32),
+    )
+    return embed_model
```
tadejsv, line 23: Make this a one liner

tests/integration/test_model_checkpoint.py
```diff
@@ -0,0 +1,52 @@
...
+
+    finetuner.fit(
+        pytorch_model,
+        epochs=1,
+        train_data=generate_fashion(num_total=1000),
+        eval_data=generate_fashion(is_testset=True, num_total=200),
+        callbacks=[BestModelCheckpoint(save_dir=tmpdir)],
+    )
+
+    assert os.listdir(tmpdir) == ['best_model_val_loss']
```
tadejsv, line 52: Remove this, it is already covered in framework-specific integration tests

tests/integration/test_model_checkpoint.py
```diff
@@ -0,0 +1,52 @@
...
+    return embed_model
+
+
+def test_both_checkpoints(pytorch_model: BaseTuner, tmpdir):
+
+    finetuner.fit(
+        pytorch_model,
+        epochs=1,
+        train_data=generate_fashion(num_total=1000),
+        eval_data=generate_fashion(is_testset=True, num_total=200),
```
tadejsv, line 32: Use the things from `conftest` here, not `generate_fashion`

finetuner/tuner/callback/training_checkpoint.py
```diff
@@ -0,0 +1,143 @@
...
+        Returns the file path for checkpoint.
+        """
+
+        file_path = os.path.join(
+            self._save_dir, f'saved_model_epoch_{tuner.state.epoch + 1:02d}'
+        )
+        return file_path
+
+    @staticmethod
+    def load_model(tuner: 'BaseTuner', fp: str):
```
tadejsv, line 121: Can we rename this to just `load` (as it loads more than the model)?

tadejsv (commented):
Final comments, then it should be ready to go

tadejsv (approved):
LGTM

JoanFM (approved):

PR #253, opened 2021-12-03
fix(test): fix flaky overfit test

makram93:
Without setting manual seed, in each test runs
- Batch sampler gives different batches
- Model fine-tuned reports different embedding because the weights initialise are always random

This causes the embeddings to be different each time and thus test is flaky

Closes issue #252

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,13 +1,27 @@
+import random as python_random
+
+import numpy as np
+import paddle
 import pytest
 import tensorflow as tf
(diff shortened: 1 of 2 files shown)


bwanglzu (approved):
LGTM! even better if we could autouse it in conftest, to make sure every test is running under same seed.

tadejsv (changes requested):
This should be a fixture

tadejsv (approved):

PR #256, opened 2021-12-06
fix: expose arguments of embed() in the evaluator

gmastrapas:


finetuner/tuner/evaluation.py
```diff
@@ -49,6 +53,16 @@ def __init__(
             match.
         :param catalog: A sequence of documents, against which the eval docs will be matched.
         :param embed_model: The embedding model to use, in order to extract document representations.
+        :param device: the computational device for `embed_model`, can be either
+            `cpu` or `cuda`.
+        :param batch_size: number of Documents in a batch for embedding
```
JoanFM, line 58: why is `batch_size` not a parameter of `embed`? why is it in the `Evaluation` constructor?

gmastrapas: It is a parameter of `embed`. The evaluator uses the embed method internally and I need a way to configure the embed behavior

JoanFM: My point is, u are telling with this design, that batch_size is part of the identity of an Evaluator. Meaning that if I want to evaluate with different batch sizes, this should change, while batch_size is just an operational parameter, but should not be vital and core to the Evaluator

gmastrapas: Ah I see, so you mean it should be  an argument of `evaluate`?

finetuner/tuner/evaluation.py
```diff
@@ -38,6 +38,10 @@ def __init__(
         eval_data: 'DocumentSequence',
         catalog: 'DocumentSequence',
         embed_model: Optional['AnyDNN'] = None,
+        device: str = 'cpu',
+        batch_size: int = 256,
+        preprocess_fn: Optional['PreprocFnType'] = None,
+        collate_fn: Optional['CollateFnType'] = None,
         distance: str = 'cosine',
         limit: int = 20,
```
maximilianwerk, line 46: limit should be right after `embed_model`. Reasoning: `limit` is a logical parameter, while the others are "technical".

PR #263, opened 2021-12-07
feat: add multiprocessing dataloading

tadejsv:
- [x] torch
- [x] paddle
- [x] ~~keras~~ - could not get keras to work with passthrough of `workers`, (and `multiprocessing`, etc.). By default Keras already behaves as if it has a (single) background process loading data - leaving it at that for now
- [x] tests (how?)

### DIFF
--- a/finetuner/tuner/__init__.py
+++ b/finetuner/tuner/__init__.py
@@ -90,7 +90,7 @@ def fit(
         will be pre-prended to this list.
     :param num_workers: Number of workers used for loading the data.
 
-        This works onlywith Pytorch and Paddle Paddle, and has no effect when using
+        This works only with Pytorch and Paddle Paddle, and has no effect when using
         a Keras model.
     """
     ft = _get_tuner_class(embed_model)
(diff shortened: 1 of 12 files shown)


finetuner/tuner/__init__.py
```diff
@@ -87,6 +88,10 @@ def fit(
         ``"cpu"`` and ``"cuda"`` (for GPU)
     :param callbacks: A list of callbacks. The progress bar callback
         will be pre-prended to this list.
+    :param num_workers: Number of workers used for loading the data.
+
+        This works onlywith Pytorch and Paddle Paddle, and has no effect when using
```
davidbp, line 93: mispelled `only with`

finetuner/tuner/base.py
```diff
@@ -176,6 +177,10 @@ def fit(
         :param batch_size: The batch size to use for training and evaluation
         :param num_items_per_class: Number of items from a single class to include in
             the batch. Only relevant for class datasets
+        :param num_workers: Number of workers used for loading the data.
+
+            This works onlywith Pytorch and Paddle Paddle, and has no effect when using
```
davidbp, line 182: same here

bwanglzu (approved):
LGTM!

bwanglzu:
FYI @gmastrapas

PR #266, opened 2021-12-08
feat: early stop callback

### LINKED ISSUE #242: Early stop callback

azayz:
closes #242


finetuner/tuner/callback/early_stopping.py
```diff
@@ -0,0 +1,110 @@
...
+                self._monitor,
+            )
+        else:
+            if self._monitor_op(current, self._best):
+                self._best = current
+                self._wait = 0
+            else:
+                self._wait += 1
+                if self._wait == self._patience:
+                    self._logger.success(
```
bwanglzu, line 105: not sure about log level, `info`?

finetuner/tuner/callback/early_stopping.py
```diff
@@ -0,0 +1,110 @@
...
+                        'Training is stopping, no improvement for {} epochs'.format(
+                            self._patience
+                        )
+                    )
+                    tuner._stop_training = True
```
bwanglzu, line 110: can you make `_stop_traning` public? to pass through different classes

bwanglzu (changes requested):
2 minor things:
1. can we add a base value (default `None`) to prevent early stopping?
2. i'm curious how it performs in a notebook, i mean this `break` will actually break the `fit`?

finetuner/tuner/callback/early_stopping.py
```diff
@@ -0,0 +1,110 @@
...
+                self._monitor_op = np.less
+                self._best = np.Inf
+
+    def on_epoch_end(self, tuner: 'BaseTuner'):
+        """
+        Called at the end of the training epoch.
+        """
+        self._check(tuner)
+        self._train_losses = []
+        self._valid_losses = []
```
gvondulong, line 75: Can you change the naming here, too? `valid_loss` sounds like there might be an invalid loss somewhere. :) 
Longer names are generally better than making the reader think.

finetuner/tuner/callback/early_stopping.py
```diff
@@ -0,0 +1,110 @@
...
+            current = np.mean(self._train_losses)
+        if current is None:
+            self._logger.warning(
+                'Can save best model only with %s available, ' 'skipping.',
+                self._monitor,
+            )
+        else:
+            if self._monitor_op(current, self._best):
+                self._best = current
+                self._wait = 0
```
gvondulong, line 101: Another naming thing. Sorry. :) 
Can we rename `_wait` to `epoch_counter`. Wait is a verb and should better be used for func names and bools.

finetuner/tuner/callback/early_stopping.py
```diff
@@ -0,0 +1,110 @@
...
+            else:
+                self._wait += 1
+                if self._wait == self._patience:
+                    self._logger.info(
+                        'Training is stopping, no improvement for {} epochs'.format(
```
gvondulong, line 106: I think it is generally best practice to use f-strings in Python 3.x. There are times to be an OG, but coding ain't one. :)

finetuner/tuner/keras/__init__.py
```diff
@@ -160,6 +160,9 @@ def fit(
         self.state = TunerState(num_epochs=epochs)
         self._trigger_callbacks('on_fit_begin')
 
+        # Check for early stopping
+        self.stop_training = False
```
gvondulong, line 164: This attribute should not be declared inside the function, but rather in the base class.

finetuner/tuner/callback/early_stopping.py
```diff
@@ -0,0 +1,110 @@
...
+            mode = 'auto'
+
+        if mode == 'min':
+            self._monitor_op = np.less
+            self._best = np.Inf
+        elif mode == 'max':
+            self._monitor_op = np.greater
+            self._best = -np.Inf
+        else:
+            if 'acc' in self._monitor:  # to adjust other metrics are added
```
gvondulong, line 62: this is not super safe. Try being more parsimonious when comparing strings.

finetuner/tuner/callback/early_stopping.py
```diff
@@ -0,0 +1,110 @@
...
+        self._valid_losses.append(tuner.state.current_loss)
+
+    def _check(self, tuner):
+        """
+        Checks if training should be stopped. If `True`
+        it stops it.
+        """
+
+        if self._monitor == 'val_loss':
+            current = np.mean(self._valid_losses)
```
gvondulong, line 90: I think the name `current` could be more descriptive, too. This is the mean epoch validation loss. The name should reflect that.

gvondulong (commented):
Good work. I left some comments, mainly about style. There is only one major thing concerning the early stopping logic. Currently, I think you are not returning the correct model.

finetuner/tuner/callback/early_stopping.py
```diff
@@ -0,0 +1,110 @@
...
+        else:
+            current = np.mean(self._train_losses)
+        if current is None:
+            self._logger.warning(
+                'Can save best model only with %s available, ' 'skipping.',
```
gvondulong, line 95: Also here, please use f-strings for string formatting.

finetuner/tuner/callback/early_stopping.py
```diff
@@ -0,0 +1,110 @@
...
+            self._best = -np.Inf
+        else:
+            if 'acc' in self._monitor:  # to adjust other metrics are added
+                self._monitor_op = np.greater
+                self._best = -np.Inf
+            else:
+                self._monitor_op = np.less
+                self._best = np.Inf
+
+    def on_epoch_end(self, tuner: 'BaseTuner'):
```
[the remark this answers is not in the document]
azayz: I'm implementing the methods of the abstract class BaseCallback so I don't think I can specify the function type hint and return at this level

finetuner/tuner/callback/early_stopping.py
azayz: I agree, changed to `validation_losses`

finetuner/tuner/callback/early_stopping.py
azayz: I agree aswell

finetuner/tuner/callback/early_stopping.py
```diff
@@ -0,0 +1,110 @@
...
+class EarlyStopping(BaseCallback):
+    """
+    Callback to stop training when a monitored metric has stopped improving.
+    A `model.fit()` training loop will check at end of every epoch whether
+    the monitered metric is no longer improving.
+    """
+
+    def __init__(
+        self,
+        monitor: str = 'val_loss',
```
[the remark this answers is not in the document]
bwanglzu: i think `monitor` is [commonly used](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/EarlyStopping), such as in tensor flow. let's keep it

finetuner/tuner/callback/early_stopping.py
```diff
@@ -0,0 +1,118 @@
...
+        mode: str = 'auto',
+        patience: int = 2,
+        min_delta: int = 0,
+        baseline: float = None,
```
bwanglzu, line 27:
```suggestion
        baseline: Optional[float] = None,
```

bwanglzu (changes requested):
in the tests, i don't think all the arguments in early stopping is tested. I know it's a bit tricky, but i think it's important to verify at least some important arguments. For example, we should capture in pytest fixture `caplog` that `... no improvements for 2 epochs` in your test.

finetuner/tuner/callback/early_stopping.py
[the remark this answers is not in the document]
gvondulong: I don't think there is anything forbidding us to document functions. But you are right. :) If you do not except any other params than the base class, you can leave the type hints. 
But it would be nice, if you explained in a few words what your implementation of the function does.

finetuner/tuner/base.py
```diff
@@ -92,6 +92,9 @@ def __init__(
         self._scheduler = None
         self._device_name = device
 
+        # Check for early stopping
+        self.stop_training = False
```
tadejsv, line 96: I would make this a private property, like all others

```suggestion
        self._stop_training = False
```

finetuner/tuner/callback/early_stopping.py
```diff
@@ -0,0 +1,118 @@
...
+    the monitered metric is no longer improving.
+    """
+
+    def __init__(
+        self,
+        monitor: str = 'val_loss',
+        mode: str = 'auto',
+        patience: int = 2,
+        min_delta: int = 0,
+        baseline: Optional[float] = None,
```
tadejsv, line 27: Add a description of this parameter in the docstring as well

finetuner/tuner/callback/early_stopping.py
```diff
@@ -0,0 +1,118 @@
...
+        it stops it.
+        """
+        if self._baseline is not None:
+            self._best = self._baseline
+
+        if self._monitor == 'val_loss':
+            current_value = np.mean(self._validation_losses)
+        else:
+            current_value = np.mean(self._train_losses)
+        if current_value is None:
```
tadejsv, line 104: You never set a default value for `current_value` - you might want to set it to `None` somewhere, otherwise this could error out

finetuner/tuner/callback/early_stopping.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        Checks if training should be stopped. If `True`
+        it stops it.
+        """
+        if self._baseline is not None:
+            self._best = self._baseline
+
+        if self._monitor == 'val_loss':
+            current_value = np.mean(self._validation_losses)
+        else:
```
tadejsv, line 102: I would be more explicit here - use an `elif` to check if it is `'train_loss'`, and in the `else` statement put the warning you have below (about the required metric not being available)

finetuner/tuner/callback/early_stopping.py
```diff
@@ -0,0 +1,118 @@
...
+    def on_val_batch_end(self, tuner: 'BaseTuner'):
+        self._validation_losses.append(tuner.state.current_loss)
+
+    def _check(self, tuner):
+        """
+        Checks if training should be stopped. If `True`
+        it stops it.
+        """
+        if self._baseline is not None:
+            self._best = self._baseline
```
tadejsv, line 98: I think this is not how baseline is used in Keras, I guess this is where you got the inspiration from
https://github.com/keras-team/keras/blob/2c48a3b38b6b6139be2da501982fd2f61d7d48fe/keras/callbacks.py#L1822-L1829

There baseline is just some lower bound on what the metric needs to be

finetuner/tuner/keras/__init__.py
```diff
@@ -184,6 +187,9 @@ def fit(
 
                 self._trigger_callbacks('on_epoch_end')
 
+                if self.stop_training:
```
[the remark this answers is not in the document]
tadejsv: @gvondulong I don't think we need to return the best model - we have the Checkpoint callback for saving. The checkpoint should then also facilitate loading the best model. The `Tuner`'s fit function is not supposed to return anything. Otherwise there is just too much coupling

tests/unit/tuner/callback/test_early_stopping.py
```diff
@@ -0,0 +1,145 @@
+import pytest
```
tadejsv, line 1: General comment: you do not need to use the `tuner.fit` function here at all, just create a tuner and then call the callback method with that tuner as argument directly. Calling the `.fit` function should only be done in an integration test - but designing one for this might not be so easy, and it might not be needed.

And current tests do not actually test the early stopping functionality - i guess because you can not so easily create a model and data that would give you the desired behavior when tuning. that's where calling callback methods directly is useful

tadejsv (changes requested):

finetuner/tuner/base.py
azayz: Bo specifically asked for this to be public in an above comment, originally it was private, I think @bwanglzu  has some use in mind for this in other modules

finetuner/tuner/callback/early_stopping.py
azayz: I'm not sure I'm following here, so the baseline is not a lower bound of what the metric should be ?

tests/unit/tuner/callback/test_early_stopping.py
azayz: I agree I changed all the tests and removed .fit

finetuner/tuner/base.py
tadejsv: Got it, let's have it public then

finetuner/tuner/callback/early_stopping.py
tadejsv: This is not resolved, the description is still not added

finetuner/tuner/callback/early_stopping.py
tadejsv: Ah I see, your implementation is actually correct

finetuner/tuner/callback/early_stopping.py
tadejsv: Not resolved

finetuner/tuner/callback/early_stopping.py
tadejsv: Not resolved

tests/unit/tuner/callback/test_early_stopping.py
```diff
@@ -0,0 +1,175 @@
...
+from finetuner.tuner.callback import EarlyStopping
+from finetuner.tuner.keras import KerasTuner
+from finetuner.tuner.paddle import PaddleTuner
+from finetuner.tuner.pytorch import PytorchTuner
+from finetuner.tuner.state import TunerState
+
+
+@pytest.fixture(scope='module')
+def pytorch_model() -> BaseTuner:
+    embed_model = torch.nn.Sequential(
```
tadejsv, line 17: Simplify this so that it is max 1 line - we need the simplest model possible. Same goes for all frameworkds

tests/unit/tuner/callback/test_early_stopping.py
```diff
@@ -0,0 +1,175 @@
...
+    checkpoint = EarlyStopping(mode=mode, monitor=monitor)
+    assert checkpoint._monitor_op == operation
+    assert checkpoint._best == best
+
+
+def test_early_stopping_pytorch(pytorch_model: BaseTuner):
+
+    tuner = PytorchTuner(
+        embed_model=pytorch_model,
+        optimizer=torch.optim.Adam(params=pytorch_model.parameters(), lr=0.001),
```
tadejsv, line 76: Custom optimizer not needed - same for all frameworks

tadejsv (commented):
Some comments, otherwise implementation/tests look good

tests/unit/tuner/callback/test_early_stopping.py
```diff
@@ -0,0 +1,152 @@
...
+from finetuner.tuner.pytorch import PytorchTuner
+from finetuner.tuner.state import TunerState
+
+
+@pytest.fixture(scope='module')
+def pytorch_model() -> BaseTuner:
+    embed_model = torch.nn.Sequential(
+        torch.nn.Linear(
+            in_features=10,
+            out_features=10,
```
tadejsv, line 20: Do this, then it will fit in one line. And you can remove the `Sequential`, in torch just `Linear` is already a valid model
```suggestion
            out_features=10
```

tests/unit/tuner/callback/test_early_stopping.py
```diff
@@ -0,0 +1,152 @@
...
+        ),
+    )
+    return embed_model
+
+
+@pytest.fixture(scope='module')
+def keras_model() -> BaseTuner:
+    embed_model = tf.keras.Sequential(
+        [
+            tf.keras.layers.Dense(10),
```
tadejsv, line 30: Same here, will be more compact (but here you can't remove sequential)
```suggestion
            tf.keras.layers.Dense(10)
```

tests/unit/tuner/callback/test_early_stopping.py
```diff
@@ -0,0 +1,152 @@
...
+    )
+    return embed_model
+
+
+@pytest.fixture(scope='module')
+def paddle_model() -> BaseTuner:
+    embed_model = paddle.nn.Sequential(
+        paddle.nn.Linear(
+            in_features=10,
+            out_features=10,
```
tadejsv, line 41: Same here 0 you can remove the `Sequential`, just `Linear` is enough 
```suggestion
            out_features=10
```

tadejsv (commented):
Some comments, looks good otherwise

tadejsv (approved):
LGTM

bwanglzu (approved):
LGTM!

PR #268, opened 2021-12-09
fix: embedding type

azayz:


bwanglzu (approved):
will find out a better solution for dtype casting, merge it to non-block all PRs.

PR #269, opened 2021-12-09
feat(tailor): vision transformer support

bwanglzu:
C644DB4F-3409-44A6-8E13-0DCA3D9FD231


bwanglzu:
will ocntinue in ee version.

PR #282, opened 2021-12-16
chore: set next version

tadejsv:


bwanglzu (approved):

JoanFM:
This seems it should be the responsibility of CD right?

tadejsv:
@JoanFM No, CD bumps the patch version - to increase minor you need to do it yourself. Currently CD creates the tags, so this is the source of truth

PR #284, opened 2021-12-17
feat: evaluator integration

gmastrapas:
Integrate the Evaluator in `finetuner.fit()`


finetuner/tuner/base.py
```diff
@@ -210,12 +220,86 @@ def _get_data_loader(
...
+    ):
+        """Embed the evaluation data and ompute the evaluation metrics"""
+
+        self.state.num_batches_query = self._get_num_batches(query_data, batch_size)
+        self.state.num_batches_index = (
+            self._get_num_batches(index_data, batch_size) if index_data else 0
+        )
+        self.state.batch_index = 0
+
+        self._trigger_callbacks('on_val_begin')
```
JoanFM, line 253: `val` or `test`? Normally in `validation` split you just control the `loss` and in `test` split you check the actual evaluation metric related to the business value

JoanFM (changes requested):

finetuner/tuner/base.py
gmastrapas: we made the decision to bypass the validation loss completely. Now it's only two phases, train and eval. In train you train and calculate the training loss, in eval you compute the search related metrics

finetuner/tuner/base.py
```diff
@@ -210,12 +220,86 @@ def _get_data_loader(
...
+                device=self._device_name,
+                batch_size=batch_size,
+                preprocess_fn=preprocess_fn,
+                collate_fn=collate_fn,
+            )
+            self._trigger_callbacks('on_val_query_batch_end')
+
+        self._trigger_callbacks('on_val_query_end')
+
+        if index_data:
```
JoanFM, line 273: what is the reason why `index_data` needs to be embedded? 

If I want to evaluate I want to evaluate against something that I am not affecting right?

gmastrapas: as the model changes during training, index data are represented differently

finetuner/tuner/base.py
JoanFM: why was this decision taken? validation step is a common practice.

But anyway, the naming is still confusing to much of the community

gmastrapas: I think it's still eval though because this step happens at the end of every epoch, not at the end of training

docs/basics/data-format.md
```diff
@@ -4,15 +4,17 @@
 Finetuner uses Jina [`Document`](https://docs.jina.ai/fundamentals/document/) as the primitive data type. In
 particular, [`DocumentArray`](https://docs.jina.ai/fundamentals/document/documentarray-api/)
 and [`DocumentArrayMemap`](https://docs.jina.ai/fundamentals/document/documentarraymemmap-api/) are the input data type
-in the high-level `finetuner.fit()` API. This means, your training dataset and evaluation dataset should be stored in `DocumentArray`
-or `DocumentArrayMemap`, where each training or evaluation instance is a `Document` object:
+in the high-level `finetuner.fit()` API. This means, your training dataset and your validation data (query/index dataset)
```
JoanFM, line 7: to me is very hard to understand what is expected in these `query` and `index` data.

finetuner/tuner/base.py
JoanFM: To me `val` reminds me of `validation` loss not about `eval`. The naming is important. You can say `on_eval_begin` not on `val_begin`.

`Validation` tends to be about loss and `evaluation` about other metrics

docs/basics/data-format.md
gmastrapas: yeah we should explain it better in the docs

bwanglzu (commented):
i'm lost, think from a user's perspective. what is a `index_data`, what is a `query_data`, how can I just do train/test split to log the validation loss on machine learning task if I only care about evaluation at the end of training?

finetuner/tuner/keras/__init__.py
```diff
@@ -97,45 +97,42 @@ def _train(self, data: KerasSequenceAdapter):
...
-        :param batch_size: The batch size to use for training and evaluation
-        :param num_items_per_class: Number of items from a single class to include in
-            the batch. Only relevant for class datasets
+            embedding model.
+        :param num_workers: Number of workers used for loading the data.
+            This works only with Pytorch and Paddle Paddle, and has no effect when using
+            a Keras model.
+        :param limit: The number of top search results to consider, when evaluating.
+        :param distance: The type of distance metric to use when matching query and index docs during evaluation,
+            available options are ``"cosine"``, ``"euclidean"`` and ``"sqeuclidean"``.
```
nan-wang, line 135:
```suggestion
            available options are ``'cosine'``, ``'euclidean'`` and ``'sqeuclidean'``.
```

finetuner/tuner/base.py
```diff
@@ -157,34 +160,41 @@ def embed_model(self) -> AnyDNN:
     def fit(
         self,
         train_data: 'DocumentSequence',
-        eval_data: Optional['DocumentSequence'] = None,
+        query_data: Optional['DocumentSequence'] = None,
+        index_data: Optional['DocumentSequence'] = None,
```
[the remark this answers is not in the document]
gmastrapas: Actually, we discussed this and ended up in the following: We expose 3 arguments, `eval_data`, `query_data`, `index_data`. Eval data have to do with the validation loss calculation and are optional. Query and index data have to do with the metrics computation and are both optional. If both are None no metrics are computed. If only index is None, metrics are computed and query data are used both for query and index. Of course a user can pass the same dataset both for eval data and query data

finetuner/tuner/callback/early_stopping.py
```diff
@@ -47,7 +48,7 @@ def __init__(
             baseline.
         :param verbose: Wheter to log score improvement events.
         """
-        self._logger = logging.getLogger('finetuner.' + self.__class__.__name__)
+        self._logger = logging.getLogger("finetuner." + self.__class__.__name__)
```
JoanFM, line 51: use single quotes to align with other jina projects

JoanFM (changes requested):

finetuner/tuner/callback/early_stopping.py
gmastrapas: oops, modified automatically by black

azayz: when using local black do: `black -S filename.py`

tadejsv (commented):
Some general comments for now

finetuner/tuner/__init__.py
```diff
@@ -52,29 +54,33 @@ def fit(
     device: str = 'cpu',
     callbacks: Optional[List['BaseCallback']] = None,
     num_workers: int = 0,
+    limit: int = 20,
+    distance: str = 'cosine',
```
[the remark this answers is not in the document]
gmastrapas: you mean the user building the evaluator and then passing it to fit? We want that? How about `eval_limit` and `eval_distance`?

finetuner/tuner/callback/best_model_checkpoint.py
```diff
@@ -29,102 +30,116 @@ def __init__(
...
             current = np.mean(self._train_losses)
+        elif self._monitor == 'val_loss':
+            current = np.mean(self._val_losses)
+        else:
+            try:
+                current = tuner.state.eval_metrics[self._monitor]
+            except KeyError:
+                current = tuner.state.eval_metrics.get(
+                    __evaluator_mean_prefix__ + self._monitor, None
+                )
```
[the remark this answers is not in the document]
gmastrapas: all metrics that are returned by the `evaluate` method start with `mean_` to indicate the averaging over all docs. Metrics inside each doc do not start with `mean_`

docs/components/tuner.md
```diff
@@ -287,7 +287,7 @@ tuner = PytorchTuner(
 )
 
 tuner.fit(
-    train_data, val_data, preprocess_fn=preprocess_fn, epochs=90, num_items_per_class=32
+    train_data, query_data, preprocess_fn=preprocess_fn, epochs=90, num_items_per_class=32
```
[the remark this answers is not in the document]
gmastrapas: yeah this is wrong. This was left from when I updated the docs when I removed the `eval_data` completely. I'll recheck the documentation

finetuner/tuner/callback/best_model_checkpoint.py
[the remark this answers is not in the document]
tadejsv: Wouldn't it make more sense that returned metrics simply have the name without any prefix?

finetuner/tuner/base.py
```diff
@@ -157,51 +152,152 @@ def fit(
...
             self._trigger_callbacks('on_keyboard_interrupt')
         except BaseException as e:
             self._trigger_callbacks('on_exception', exception=e)
             raise
 
+    @staticmethod
+    def _get_num_batches(docs: 'DocumentSequence', batch_size: int) -> int:
+        """Get the number of batches from a document sequence."""
+        n = len(docs)
+        return n // batch_size + 1 * (n % batch_size)
```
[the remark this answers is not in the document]
gmastrapas: Yes you are right, I wanted to write `n // batch_size + 1 * (n % batch_size > 0)`

finetuner/tuner/callback/best_model_checkpoint.py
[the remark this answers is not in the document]
gmastrapas: Probably yes. I'll simplify it for now so we don't have this awkward handling

finetuner/tuner/__init__.py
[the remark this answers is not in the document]
tadejsv: I guess at one point we will also allow the user to specify which metrics to include, and maybe other evaluator kwargs. So then the question is do we copy them all to tuner, or do we abstract them away in the Evaluator instance. What do you think?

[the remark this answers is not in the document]
gmastrapas: I think since we decided to integrate the evaluator we should go all the way. Which means expose the eval related args in fit. As a user I don't like having to build the evaluator before calling fit. We also lose the one-line execution of `finetuner.fit`

tadejsv (commented):
So, I was thinking a bit about this - I think it would make more sense to include evaluator as a callback. The reason is that there is no strong integration with the rest of the training loop, and the hooks that we create for it are unlikely to be used by any other callback other than the progress bar.

This would ease the load on the already long list of arguments passed to finetuner. The callback would need its own progress bar (which is not very difficult to do), and communication with other callbacks would be through writing in the `state` (just like now). The only thing here is that the user will need to take care when passing callbacks to put the evaluator callback before the callbacks that rely on metrics.

finetuner/tuner/__init__.py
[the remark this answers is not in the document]
tadejsv: I thought a bit about this - maybe it makes more sense to implement this as a callback. I wrote more in detail in a comment below.

gmastrapas:
@tadejsv if we do it as a callback the user will need to pass many arguments of fit in the callback. Example
```
callbacks=[
    Evaluation(
        query_data,
        index_data,
        embed_model,
        batch_size=batch_size,
        device=device,
        preprocess_fn=preprocess_fn,
        collate_fn=collate_fn,
    )
]
```
I don't think this is very usable. Another option is to add these arguments to the state, so that the callback can retrieve them, but having these in the state doesn't make much sense to me. What do you think?

Also what happens if the user uses `to_embedding_model=True`? The model that should be used for evaluation is not the same that the user pass in fit

tadejsv:
@gmastrapas 
> I don't think this is very usable. Another option is to add these arguments to the state, so that the callback can retrieve them, but having these in the state doesn't make much sense to me. What do you think?

So of the arguments you have shown, only `query_data` and `index_data` (and metrics related args) would need to be passed to the callback. For the model and the rest, we can set them as a (private) property of the tuner (I agree they don't really belong to the state), and the callback can get them from there - it has access to `tuner`.

gmastrapas:
> we can set them as a (private) property of the tuner (I agree they don't really belong to the state), and the callback can get them from there - it has access to tuner.

This makes perfect sense, I'll proceed this way

docs/components/tuner/evaluation.md
```diff
@@ -0,0 +1,61 @@
...
+or integrated in the training loop via the [evaluation callback](#using-the-evaluation-callback).
+
+
+## Using the evaluator
+
+The evaluator can be used standalone in order to compute the evaluation metrics on a sequence
+of documents:
+```python
+from finetuner.tuner.evaluation import Evaluator
+from jina import DocumentArray
```
bwanglzu, line 15:
```suggestion
from docarray import DocumentArray
```

docs/components/tuner/evaluation.md
```diff
@@ -0,0 +1,61 @@
...
+
+If an embedding model is given, the query docs are embedded, otherwise they are assumed to carry
+representations.
+
+The `index_data` (or catalog) is an optional argument that defines the dataset against which the
+query docs are matched. If not provided, query docs are matched against themselves.
+
+The `evaluate()` method returns the computed metrics as a dictionary mapping metric names to values.
+Common IR metrics, like precision, recall, average precision, dcg, ndcg and reciprocal rank are
+included.
```
bwanglzu, line 38: can you make a list here?

docs/components/tuner/evaluation.md
```diff
@@ -0,0 +1,61 @@
...
+from finetuner.tuner.callback import Evaluation
+from jina import DocumentArray
+
+query_data = DocumentArray(...)
+index_data = DocumentArray(...)
+
+finetuner.fit(
+    ...,
+    callbacks=[Evaluation(query_data, index_data)],
+)
```
bwanglzu, line 59: it's a bit unclear for me, how to compose `query_data` and `index_data`, and how to evaluate on certain metrics, such as mAP?

finetuner/tuner/__init__.py
```diff
@@ -56,25 +56,25 @@ def fit(
...
         items into a batch. Should accept a list with the content of each item,
         and output a tensor (or a list/dict of tensors) that feed directly into the
-        embedding model
-    :param epochs: Number of epochs to train the model
-    :param batch_size: The batch size to use for training and evaluation
+        embedding model.
+    :param epochs: Number of epochs to train the model.
+    :param batch_size: The batch size to use for training and evaluation.
+    :param num_items_per_class: Number of items from a single class to include in
+        the batch. Only relevant for class datasets.
```
bwanglzu, line 73: :class:`ClassDataset`

finetuner/tuner/callback/evaluation.py
```diff
@@ -0,0 +1,132 @@
...
+class Evaluation(BaseCallback):
+    """
+    A callback that uses the Evaluator to calculate IR metrics at the end of each epoch. When used
+    with other callbacks that rely on metrics, like checkpoints and logging, this callback should be
+    defined first, so that it precedes in execution.
+    """
+
+    def __init__(
+        self,
+        query_data: Optional['DocumentSequence'],
```
bwanglzu, line 23:
```suggestion
        query_data: Optional['DocumentSequence'] = None,
```

finetuner/tuner/callback/evaluation.py
```diff
@@ -0,0 +1,132 @@
...
+from ..dataset.helper import batch_document_sequence
+from ..evaluation import Evaluator
+from .base import BaseCallback
+
+if TYPE_CHECKING:
+    from ...helper import DocumentSequence
+    from ..base import BaseTuner
+
+
+class Evaluation(BaseCallback):
```
bwanglzu, line 14: i suggest we give a more precise naming, to clarify when use it standalone or as callback, such as name it `EvaluationCallback`.

finetuner/tuner/dataset/helper.py
```diff
@@ -0,0 +1,15 @@
...
+if TYPE_CHECKING:
+    from ...helper import DocumentSequence
+
+
+def batch_document_sequence(docs: 'DocumentSequence', size: int):
+    """Batch a document sequence into DocumentArray batches"""
+    iterable: Iterator[Any] = iter(docs)
+    return iter(
+        lambda: DocumentArray(list(itertools.islice(iterable, size))), DocumentArray()
+    )
```
bwanglzu, line 15: i don't understand this function, why not `da.batch`? https://docs.jina.ai/fundamentals/document/documentarray-api/?highlight=da%20batch#batching

bwanglzu (changes requested):
left some comments

docs/basics/data-format.md
```diff
@@ -4,15 +4,17 @@
...
+should be stored in `DocumentArray` or `DocumentArrayMemap`, where each training, query or index instance is a `Document`
+object:
 
 ```python
 import finetuner
 
 finetuner.fit(model,
               train_data=...,
-              eval_data=...)
+              query_data=...,
```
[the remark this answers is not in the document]
gmastrapas: I think these two are outdated

docs/components/tuner/evaluation.md
gmastrapas: selecting specific metrics is not supported currently. query and index data are standard class or session format doc arrays

docs/components/tuner/evaluation.md
gmastrapas: sure

finetuner/tuner/dataset/helper.py
gmastrapas: because document sequence can be a list of documents

docs/components/tuner/evaluation.md
bwanglzu: i see

finetuner/tuner/dataset/helper.py
bwanglzu: how about convert it into a DA beforehead? I forgot why do we support list of Documents, seems now everyone is using the DA as input to finetuner @tadejsv

bwanglzu (changes requested):
minor comment, besides, can you try if the callback progress bar looks correct in notebook?

finetuner/tuner/dataset/helper.py
bwanglzu: okay

tadejsv: I agree, let's wait for docarray to be released, then we can make this change

gmastrapas:
> minor comment, besides, can you try if the callback progress bar looks correct in notebook?

@bwanglzu It works if I install `ipywidgets`. A warning however from the same package is constantly being refreshed with the progress bar

finetuner/tuner/state.py
```diff
@@ -1,10 +1,14 @@
...
 
+from rich.progress import Progress
+
 
 @dataclass
 class TunerState:
-    """A data container representing containing information on the current run."""
+    """A data container representing information on the current run."""
+
+    progress_bar: Progress = Progress()
```
tadejsv, line 11: Hm, why do we need this? Does creating a second (independent) progress bar cause any issues?

docs/components/tuner/evaluation.md
```diff
@@ -0,0 +1,72 @@
...
+from finetuner.tuner.evaluation import Evaluator
+from docarray import DocumentArray
+
+query_data = DocumentArray(...)
+index_data = DocumentArray(...)
+embed_model = ...
+
+evaluator = Evaluator(query_data, index_data, embed_model)
+
+metrics = evaluator.evaluate()
```
tadejsv, line 23: Is it possible that we provide an actual working example here? I think it would be more informative

docs/components/tuner/callbacks.md
```diff
@@ -2,12 +2,12 @@
 
 Callbacks offer a way to integrate various auxiliary tasks into the training loop. We offer built-in callbacks for some common tasks, such as
 - Showing a progress bar (which is shown by default)
-- [Tracking experiements](#experiement-tracking)
+- [Tracking experiments](#experiment-tracking)
 - [Checkpoint training progress](#checkpoints)
```
tadejsv, line 7: I would add something about evaluation callback here as well

finetuner/tuner/evaluation.py
```diff
@@ -17,88 +18,104 @@
...
-    "r_precision": r_precision,
-    "precision": precision_at_k,
-    "recall": recall_at_k,
-    "f1score": f1_score_at_k,
-    "average_precision": average_precision,
-    "hit": hit_at_k,
-    "reciprocal_rank": reciprocal_rank,
-    "dcg": dcg_at_k,
-    "ndcg": ndcg_at_k,
+    'r_precision': r_precision,
```
tadejsv, line 25: One thing here - wasn't it also planned to not rely on functions from `docarray.math.evaluation` and instead use methods of `DocumentArray`?

docs/components/tuner/evaluation.md
gmastrapas: sure I'll add an actual example instead of dots

finetuner/tuner/evaluation.py
gmastrapas: yes indeeed, but it's blocked by a change that needs to be done in docarray. I opened a separate issue for this

tadejsv: Okay, got it

bwanglzu (approved):
LGTM! Before release, we might give a end-to-end tutorial which includes evaluator as callback.

docs/components/tuner/evaluation.md
```diff
@@ -0,0 +1,190 @@
...
+        
+        return torch.tensor(embeddings)
+
+```
+
+Now let's create some example data. We will divide into 2 sets, the evaluation data and the index data. The
+evaluation data are docs with ids from 0 to 9. The index data are docs with ids from 10 to 19. As content, for
+each doc we use `doc.blob = np.array([doc.id % 10])`. Same content yields the same embedding, so for each
+query doc we set as target match the corresponding doc from the index data that has the same content:
+```python
```
tadejsv, line 79: Can you use two tabs for Session/Class dataset demonstration?

docs/components/tuner/evaluation.md
```diff
@@ -0,0 +1,190 @@
...
+        input_shape: (bs, 1)
+        output_shape: (bs, 10)
+        """
+        batch_size = inputs.size()[0]
+        embeddings = []
+        for i in range(batch_size):
+            idx = inputs[i][0]
+            embeddings.append([idx] * 10)
+        
+        return torch.tensor(embeddings)
```
tadejsv, line 71: I think we can use `Tensor.repeat` or something similar to make this nicer and more compact. Also, I would not make `forward` a static method.

finetuner/tuner/state.py
tadejsv: Okay, then let's make the progress bar an attribute of the `Tuner` directly

tadejsv (commented):
Looks good now, some final comments

tadejsv (approved):
LGTM!

PR #286, opened 2021-12-20
fix: handle exceptions in callbacks

tadejsv:
This adds a new `on_exception` hook to the callbacks. It is used in progress bar to properly shut it down if an exception occurs.

### DIFF
--- a/finetuner/tuner/base.py
+++ b/finetuner/tuner/base.py
@@ -143,17 +143,16 @@ class BaseTuner(abc.ABC, Generic[AnyDNN, AnyDataLoader, AnyOptimizer, AnySchedul
     def _default_configure_optimizer(self, model: AnyDNN) -> AnyOptimizer:
         """Get the default optimizer (Adam), if none was provided by user."""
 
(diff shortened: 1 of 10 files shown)


bwanglzu (approved):
LGTM!

tadejsv:
@bwanglzu i will add some changes first

finetuner/tuner/base.py
```diff
@@ -186,6 +185,20 @@ def fit(
...
+                train_data,
+                eval_data,
+                epochs,
+                batch_size,
+                num_items_per_class,
+                preprocess_fn,
+                collate_fn,
+                num_workers,
+            )
+        except BaseException as e:
```
JoanFM, line 199: why not simply Exception?

tadejsv: I think `BaseException` is the base class of `Exception`, and some exception directly subclass `BaseException`

bwanglzu (approved):
LGTM, tested in notebook, worked!

PR #303, opened 2021-12-22
fix(tuner): logging

tadejsv:
This makes logging play nicer with the progress bar - it is not interrupted, logging instead is shown above it.

I also remove any dependency on jina logger.

The new logging setup is rich-compatible, and tries to be non-intrusive - only the settings for `finetuner` logger are changed, root and other loggers are not affected.

Here's how the new logging looks like
logg_new


finetuner/toydata.py
```diff
@@ -86,24 +86,21 @@ def _download_qa_data(
...
-            if not os.path.exists(v['filename']):
-                urllib.request.urlretrieve(
-                    v['url'], v['filename'], reporthook=lambda *x: t.update(0.01)
-                )
-
-        with open(targets['covid-csv']['filename']) as fp:
-            lines = csv.DictReader(fp)
-            for idx, value in enumerate(lines):
-                if is_testset is None:
+    for k, v in track(targets.items(), 'Downloading chatbot data'):
```
JoanFM, line 89: ignore k maybe?

tadejsv: sure

bwanglzu:
does it work on notebook?

finetuner/tuner/callback/early_stopping.py
```diff
@@ -23,6 +23,7 @@ def __init__(
         patience: int = 2,
         min_delta: int = 0,
         baseline: Optional[float] = None,
+        verbose: bool = False,
```
bwanglzu, line 26: docstring is not updated here

bwanglzu (changes requested):
minor comments

tadejsv:
@bwanglzu Works similar as before, here's what I see at the end of training. The only thing is that there is still the annoying "blinking" warning message, but that's the same as before.

image

bwanglzu (approved):
LGTM!

PR #305, opened 2021-12-23
chore: CI improvements

tadejsv:
- [x] Move black, flake8, isort configuration to `pyproject.toml` and `setup.cfg`
- [x] Update CI and pre-commit hooks


.github/workflows/ci.yml
```diff
@@ -4,86 +4,48 @@ on:
...
-          CHANGED_FILES: ${{ steps.file_changes.outputs.added_modified }}
+      - name: Install requirements
+        run: pip install -r .github/requirements-test.txt
+      - name: check isort
+        run: isort --check .
 
   prep-testbed:
     runs-on: ubuntu-latest
-    needs: [commit-lint, lint-flake-8, check-black]
+    needs: [lint-flake-8, check-black, check-isort]
```
bwanglzu, line 48: i think a general better thing to do is to only apply changes on "updated files within the PR", not all files.

bwanglzu (commented):
quick comment, will continue a bit later

.github/workflows/ci.yml
tadejsv: If you start from a state where all checks pass, it is the same. Running time is negligible (<1s for all), so I don't see a good reason why we need it - it just introduces an extra step in CI

bwanglzu (approved):
LGTM!

PR #308, opened 2021-12-23
feat(tuner): onnx model dump load 280

gvondulong:
This pr adds and test export to onnx from pytorch, keras and paddle


finetuner/tuner/onnx.py
```diff
@@ -0,0 +1,158 @@
...
+)
+
+from finetuner.helper import get_framework
+
+
+def to_onnx(
+    embed_model: 'AnyDNN',
+    path: str,
+    input_shape: Union[Tuple[int], List[int]],
+    batch_size: Optional[Union[int, None]] = None,
```
gmastrapas, line 19: I think `batch_size` should not be an argument here. It's an extra parameter for the user to consider and it doesn't offer anything

finetuner/tuner/onnx.py
```diff
@@ -0,0 +1,158 @@
...
+    opset_version: int = 11,
+) -> None:
+    """Func that converts a given model in paddle, torch or keras,
+    and converts it to the ONNX format
```
gmastrapas, line 23: syntax

tests/unit/tuner/test_onnx.py
```diff
@@ -0,0 +1,288 @@
...
+    session = onnxruntime.InferenceSession(temp_onnx_file)
+
+    model.eval()
+    y_original = model(torch.Tensor(x)).detach().numpy()
+    y_exported = session.run(None, {session.get_inputs()[0].name: x})[0]
+
+    np.testing.assert_allclose(y_original, y_exported, rtol=1e-03, atol=1e-05)
+
+
+def test_pytorch_lstm_to_onnx(get_pytorch_lstm_model):
```
gvondulong, line 203: @gmastrapas, contrary to what we spoke about, concerning putting the onnx evaluation into the onnx.py file, I put it into the test file. 
Even though it would be nice to have such functionality, as a user, the different types of models e.g. LSTM have outputs different from tensors. In case of lstm, it outputs the hidden state and the memory vector. 
We would need to handle all output types for all possible types of models. 
IMHO, that would be a little over-engineered and in the end it is only a few lines of code to evaluate the export, as a user.

finetuner/tuner/onnx.py
```diff
@@ -0,0 +1,158 @@
...
+    :param input_shape: Input shape of embedding model
+    :param batch_size: The batch size the model was trained with. This
+      is only required when converting pyTorch model to onnx. In other
+      frameworks, the batch size does not need to be made specific
+    :param opset_version: ONNX opset version in which to register
+    """
+    if isinstance(input_shape, tuple):
+        input_shape = list(input_shape)
+
+    if not path.endswith(".onnx"):
```
gmastrapas, line 35: use single quotes

finetuner/tuner/onnx.py
```diff
@@ -0,0 +1,158 @@
...
+        return {
+            'torch': _to_onnx_torch,
+            'keras': _to_onnx_keras,
+            'paddle': _to_onnx_paddle,
+        }[framework_name](embed_model, path, input_shape, opset_version)
+
+    fm = get_framework(embed_model)
+
+    # Get framework-specific func to register model in ONNX
+    _parse_and_apply_to_onnx_func(fm)
```
gmastrapas, line 49: why not `getattr(globals())`. It's a single line

finetuner/tuner/onnx.py
```diff
@@ -0,0 +1,158 @@
...
+    :param embed_model: Embedding model to register in ONNX
+    :param path: Patch where to register ONNX model to
+    :param input_shape: Embedding model input shape
+    :param batch_size: The batch size the model was trained with
+    :param opset_version: ONNX opset version in which to register
+    """
+
+    import torch
+
+    embed_model.eval()
```
gmastrapas, line 81: if model is previously in training mode, it should stay that way

requirements.txt
```diff
@@ -1 +1,4 @@
-.
\ No newline at end of file
+onnx==1.9.0
```
gmastrapas, line 1: I don't think we should add these in requirements.txt. We should only try and import them in the functions that they are needed and raise errors if not found. e.g.
```
The onnx package is required to export a model in ONNX. Try 'pip install onnx'
```

tests/unit/tuner/test_onnx.py
```diff
@@ -0,0 +1,288 @@
...
+from pathlib import Path
+from typing import Any, Tuple
+
+from tensorflow import keras
+
+from finetuner.helper import get_framework
+from finetuner.tuner.onnx import to_onnx
+
+
+def validate_onnx_export(
```
gmastrapas, line 18: I really think this should be in our public API

tests/unit/tuner/test_onnx.py
```diff
@@ -0,0 +1,288 @@
...
+
+    if is_training_before:
+        embed_model.train()
+
+    np.testing.assert_allclose(y_original, y_exported, rtol=1e-03, atol=1e-05)
+
+
+@pytest.fixture
+def get_pytorch_linear_model():
+    def _get_model(input_shape):
```
gmastrapas, line 78: why not a single function? non pytest.fixture

finetuner/tuner/onnx.py
```diff
@@ -0,0 +1,158 @@
...
+    Union,
+)
+
+from finetuner.helper import get_framework
+
+
+def to_onnx(
+    embed_model: 'AnyDNN',
+    path: str,
+    input_shape: Union[Tuple[int], List[int]],
```
gmastrapas, line 18: `Sequence[int]`?

finetuner/tuner/onnx.py
gmastrapas, line 19: `Optional[int]`

finetuner/tuner/onnx.py
```diff
@@ -0,0 +1,158 @@
...
+) -> None:
+    """Convert a paddle embedding model to the ONNX format
+    :param embed_model: Embedding model to register in ONNX
+    :param path: Patch where to register ONNX model to
+    :param input_shape: Embedding model input shape
+    :param opset_version: ONNX opset version in which to register
+    """
+
+    # Removing onnx extension as paddle adds it automatically
+    if path.endswith(".onnx"):
```
gmastrapas, line 144: single quotes

gmastrapas (changes requested):
Left some comments

tests/unit/tuner/test_onnx.py
gmastrapas: I think we can stick to embedding models, i.e. assume that all models return a batch of vectors `(batch_size, dim)`

finetuner/tuner/onnx.py
```diff
@@ -0,0 +1,207 @@
...
+import numpy as np
+import onnxruntime
+
+from finetuner.helper import AnyDNN
+from finetuner.helper import get_framework
+
+
+def to_onnx(
+    embed_model: AnyDNN,
+    path: str,
```
bwanglzu, line 11: i suggest we use `model_path` here, consistent with what we have in `def save(..)`. (see tuner/__init__.py)

finetuner/tuner/onnx.py
gvondulong: Can you elaborate?

finetuner/tuner/onnx.py
```diff
@@ -0,0 +1,206 @@
...
+    :param input_shape: Embedding model input shape
+    :param batch_size: The batch size during export
+    :param opset_version: ONNX opset version in which to register
+    """
+
+    import torch
+
+    embed_model.eval()
+
+    x = torch.randn([batch_size] + input_shape, requires_grad=True, dtype=torch.float32)
```
bwanglzu, line 73: why we stick with `float32` as dtype? does it work for different kind of models? we have `input_dtype` as `float32` as default in tailor, i think we need to have some flexibility here. see [this](https://github.com/jina-ai/finetuner/blob/main/finetuner/tailor/base.py#L13) link

finetuner/tuner/onnx.py
```diff
@@ -0,0 +1,206 @@
...
+
+    # Removing onnx extension as paddle adds it automatically
+    if path.endswith(".onnx"):
+        path = path[:-5]
+
+    import paddle
+    from paddle.static import InputSpec
+
+    shape = [None] + list(input_shape)
+    x_spec = InputSpec(shape, 'float32', 'input')
```
bwanglzu, line 141: same, the `input_dtype` should be configurable

finetuner/tuner/onnx.py
```diff
@@ -0,0 +1,206 @@
...
+) -> None:
+    """
+    Test an exported model by comparing the outputs of the original and the exported model
+    against the same input.
+    :param embed_model: The original embedding model. Can be either a PyTorch module,
+        a Keras model or a PaddlePaddle layer.
+    :param export_path: The path where the exported model is stored.
+    :param input_shape: The model's expected input shape, without the batch axis.
+    """
+    fm = get_framework(embed_model)
```
bwanglzu, line 165: i think you can import `onxxruntime` here

finetuner/tuner/onnx.py
gmastrapas:
```
"""Convert a PyTorch, Keras or PaddlePaddle embedding model to the ONNX format"""
```

bwanglzu (changes requested):
some comments

bwanglzu:
can we also make sure some model trained on gpu can be exported correctly as well?

tests/unit/tuner/test_onnx.py
gvondulong: Solved in offline conversation. 
The evaluation is move back as George suggested, since we agreed it would be save to assume, that the embed model creates only one tensor as embedding for new.

finetuner/tuner/onnx.py
tadejsv: But in tailor you don't even use that property, or am i wrong?

finetuner/tuner/onnx.py
gvondulong: Good point. I changed it.

finetuner/tuner/onnx.py
tadejsv: Why not just `if/elif`

requirements.txt
tadejsv: I agree

finetuner/tuner/onnx.py
gvondulong: Changed it here.

finetuner/tuner/onnx.py
gvondulong: I changed the approach using `globals`, because I like to make it more explicit. In the end, I am super agnostic here. globals, if/else, my way are all fine to me. It all kinda is the same. :) 
I will change it to if/else, since it might be the most basic and easy way of doing it. 
If you have objections, feel free to raise them. Otherwise I am going to wrap this up?!

finetuner/tuner/onnx.py
bwanglzu: we use, actually for most of the models `float32` works, lstm seems does not (with an `Embedding` layer).

tests/unit/tuner/test_onnx.py
```diff
@@ -0,0 +1,217 @@
...
+def get_paddle_lstm_model():
+    def _get_model():
+        return paddle.nn.Sequential(paddle.nn.LSTM(32, 2))
+
+    return _get_model
+
+
+def test_pytorch_to_onnx(get_pytorch_linear_model):
+    INPUT_SHAPE = [32]
+    temp_onnx_file = str(Path(tempfile.tempdir) / 'finetuned.onnx')
```
bwanglzu, line 108: there is a pytest fixture for this https://docs.pytest.org/en/6.2.x/tmpdir.html

tests/unit/tuner/test_onnx.py
```diff
@@ -0,0 +1,217 @@
...
+def paddle_conv_model():
+    return paddle.nn.Sequential(paddle.nn.Conv2D(3, 3, 2, stride=1))
+
+
+@pytest.fixture
+def get_paddle_lstm_model():
+    def _get_model():
+        return paddle.nn.Sequential(paddle.nn.LSTM(32, 2))
+
+    return _get_model
```
bwanglzu, line 103: for all these fixtures, if you can copy fixtures from `test/unit/tailor/test_torch.py` (and keras, paddle), and move it into `conftest` will be even better. (feel free to not do it, but we need to clean this up)

bwanglzu (commented):
some comments in tests

tests/unit/tuner/test_onnx.py
bwanglzu: please ignore this comment, I'll create a common model fixture in projection head PR

tadejsv (commented):
Minor comments, looks good in general

finetuner/tuner/onnx.py
JoanFM: I am curious, why is this even needed to export a model.

JoanFM: It is weird that to export a model I need to pass a batch_size as input

JoanFM (dismissed):

requirements.txt
JoanFM: yes, makes sense. Try to mimic the ImportExtensions behavior from jina core. But copy paste and adapt, do not add dependency

finetuner/tuner/onnx.py
tadejsv: Creating an example input is needed for pytorch to trace the model for exporting - but we could have a single fixed value for batch size, since it is set as dynamic anyway

tests/unit/tuner/test_onnx.py
gvondulong: Great! Thanks!

finetuner/tuner/onnx.py
JoanFM: at least yes, u shouild have a single squeezed input

JoanFM: But let's not expose `batch-size`

gvondulong: I've added the param

bwanglzu (approved):
LGTM! Better if you can resolve threads related to `requirements.txt`.

bwanglzu:
@gmastrapas review again?

finetuner/tuner/onnx.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+def to_onnx(
+    embed_model: AnyDNN,
+    model_path: str,
+    input_shape: Sequence[int],
+    opset_version: int = 11,
+    input_type: str = 'float32',
+) -> None:
+    """Func that converts a given model in PaddlePaddle, PyTorch or Keras,
```
gmastrapas, line 16: Can we change this docstring?

finetuner/tuner/onnx.py
```diff
@@ -0,0 +1,229 @@
...
+        input_spec=[x_spec],
+        opset_version=opset_version,
+    )
+
+
+def validate_onnx_export(
+    embed_model: AnyDNN,
+    export_path: str,
+    input_shape: Tuple[int, ...],
+    batch_size: int = 8,
```
gmastrapas, line 176: Remove batch size from here as well?

requirements.txt
gmastrapas: will we keep these in the requirements file?

gmastrapas (changes requested):

finetuner/tuner/onnx.py
gvondulong: Would you suggest hard coding it below?

gmastrapas (changes requested):

gmastrapas (approved):
Looks good 🚀

PR #312, opened 2021-12-31
docs: add checkpoints to documentation

tadejsv:


docs/components/tuner/callbacks.md
```diff
@@ -32,6 +33,42 @@ tuner = PytorchTuner(..., callbacks=[logger])
...
+```python
+from finetuner.tuner.callback import TrainingCheckpoint
+from finetuner.tuner.pytorch import PytorchTuner
+
+checkpoint = TrainingCheckpoint('checkpoints')
+
+tuner = PytorchTuner(..., callbacks=[checkpoint])
+
+# Afterwards, load tuner from the saved theckpoint
+TrainingCheckpoint.load(tuner, 'checkpoints/saved_model_epoch_10')
```
bwanglzu, line 55: is it correct that the `saved_model_epoch_10` is generated automatically?

docs/components/tuner/callbacks.md
```diff
@@ -32,6 +33,42 @@ tuner = PytorchTuner(..., callbacks=[logger])
...
+```
+
+For the {class}`~finetuner.tuner.callback.best_model_checkpoint.BestModelCheckpoint`, you would also add it to `callbacks`, and later you could load the model from it
```
bwanglzu, line 58:
```suggestion
For the {class}`~finetuner.tuner.callback.best_model_checkpoint.BestModelCheckpoint`, you would also add it to `callbacks`, and later you could load the model from it.
```

bwanglzu (commented):
minor comments

docs/components/tuner/callbacks.md
tadejsv: It's what a checkpoint saved in epoch 10 would look like, yes. I can try adding an example for tf/keras as well

tadejsv: Actually, looking at this - it would be exactly the same for keras/paddle (just `KerasTuner` instead of `PytorchTuner`), so here I would not add the examples for now

bwanglzu (approved):
LGTM!

PR #316, opened 2022-01-04
feat(tuner): add default projection head for ssl

bwanglzu:
`ProjectionHead`, first proposed in SimCLR paper, is adding a 2-3 layer MLP on top of the embedding model, this trick is able to bootstrap the representation learning. This has became a common practice in contrastive representation learning.

This PR will:
1. create a simple MLP head in tailor, replace the old `bottleneck_net`.
2. add tests.
3. restructure the tailor tests, now fixtures of tailor moved out from the tailor to conftest.

- [x] pytorch
- [x] paddle
- [x] keras


finetuner/tuner/paddle/__init__.py
```diff
@@ -232,3 +232,45 @@ def get_device(device: str):
...
+        self.head_layers = nn.LayerList()
+        is_last_layer = False
+        for idx in range(num_layers):
+            if idx == num_layers - 1:
+                is_last_layer = True
+            if not is_last_layer:
+                self.head_layers.append(
+                    nn.Linear(
+                        in_features=in_features,
+                        out_features=in_features,
```
JoanFM, line 256: Why not to let the user specify the architecture? It seems like a big assumptiuon that all the layers are `in_feat, in_feat, in_feat` until the last layer

bwanglzu: i think there is not need, since this simple MLP trick works to improve SSL results. everyone is using Dense + BN + RELU combinations 2 - 3 layers, and it's a hidden head `_ProjectionHead` user will not notice

tests/unit/conftest.py
```diff
@@ -0,0 +1,265 @@
...
+def torch_dense_model():
+    return torch.nn.Sequential(
+        torch.nn.Linear(in_features=128, out_features=128),
+        torch.nn.ReLU(),
+        torch.nn.Linear(in_features=128, out_features=64),
+        torch.nn.ReLU(),
+        torch.nn.Linear(in_features=64, out_features=32),
+        torch.nn.ReLU(),
+        torch.nn.Linear(in_features=32, out_features=10),
+        torch.nn.Softmax(),
```
tadejsv, line 17: What are we trying to acheive by having so many layers in a test model?

bwanglzu: reuse of the tailor fixtures to make sure different models works when attach projection head. @tadejsv

finetuner/tuner/paddle/__init__.py
JoanFM: I think that the interface would be as simple and the assumptions are way less.

I feel we are doing a lot of assumptions here

bwanglzu: do you mean that we fix number of layers and output dimension? @JoanFM

JoanFM: u assume that the MLP is in_features x  num_layers x output_dim.

It seems weird that dimensionality is reduced slowly as layers are added

bwanglzu: i understand, it's tricky because we don't know the output dimension of the embedding model. We could "smooth", but my general impression is since it's a small MLP, it's not necessary

numb3r3:
> chop the projection head once training is finished

@bwanglzu I'm sorry to argue this trick.I think we need some experiments to see whether it is reasonable to chop the projection layer for the retrieval task.  

In related papers, they declared that the projection layer resulted in degraded performance on downstream tasks (i.e., classification task). However, they did not evaluate the performance of the retrieval task.

bwanglzu:
@numb3r3 just had a discussion with @tadejsv , we'll keep the projection head as `bottleneck_layer`, so we don't remove it anymore.  It will be used as a tailor module.

Is it sounds reasonable to you?

finetuner/tailor/keras/projection_head.py
```diff
@@ -0,0 +1,41 @@
...
+
+class ProjectionHead(tf.keras.layers.Layer):
+    """Projection head used internally for self-supervised training.
+    It is (by default) a simple 3-layer MLP to be attached on top of embedding model only for training purpose.
+    After training, it should be cut-out from the embedding model.
+    """
+
+    EPSILON = 1e-5
+
+    def __init__(self, in_features: int, output_dim: int = 128, num_layers: int = 3):
```
gvondulong, line 12: This looks totally correct. 
I think this could be made simpler w/o any branching or the flag variable `is_last_layer`. 
You could simply do: 
```
# For each but the last layer 
for idx in range(num_layers - 1): 
    self.head_layers.append(...) 
    self.head_layers.append(nn.Relu())
# Then append last layer
self.head_layers.append(...)
```
I think this makes it easier to read and reason about, but It is just a style thing, so feel free to reject it.

finetuner/tailor/keras/__init__.py
```diff
@@ -83,7 +83,7 @@ def to_embedding_model(
             will be removed. When set to ``None``, then the last layer listed in :py:attr:`.embedding_layers` will be used.
             To see all available names you can check ``name`` field of :py:attr:`.embedding_layers`.
         :param freeze: if set as True, will freeze all layers before :py:`attr`:`layer_name`. If set as list of str, will freeze layers by names.
-        :param bottleneck_net: Attach a bottleneck net at the end of model, this module should always trainable.
```
gvondulong, line 86: "... should always `be` trainable"

gvondulong:
Looks good! Just two comments. :)

finetuner/tailor/keras/projection_head.py
bwanglzu: yes! this is something I'm updating now, thanks!

finetuner/tailor/pytorch/projection_head.py
```diff
@@ -0,0 +1,30 @@
+import torch.nn as nn
+
+
+class ProjectionHead(nn.Module):
+    """Projection head used internally for self-supervised training.
+    It is (by default) a simple 3-layer MLP to be attached on top of embedding model only for training purpose.
+    After training, it should be cut-out from the embedding model.
+    """
+
+    EPSILON = 1e-5
```
tadejsv, line 10: Any reason we are specifying this manually instead of leaving it at default?

tadejsv (commented):
Looks good in general.

finetuner/tailor/pytorch/projection_head.py
bwanglzu: different frameworks might have different values, i want to set them as the same @tadejsv

tadejsv: sounds good

tadejsv (commented):
LGTM

gvondulong (approved):
lgtm

PR #326, opened 2022-01-07
feat: add NTXent loss

tadejsv:
TODO

- [x] Mention loss in documentation
- [x] GPU tests for loss


finetuner/tuner/pytorch/losses.py
```diff
@@ -174,3 +174,46 @@ def get_default_miner(
...
+        :param temerature: The temperature parameter
+        """
+        super().__init__()
+
+        self.temperature = temperature
+
+    def forward(self, embeddings: torch.Tensor, labels: torch.Tensor) -> torch.Tensor:
+        """Compute the loss.
+
+        :param embeddings: An ``[N, d]`` tensor of embeddings.
```
JoanFM, line 199: If u compute cross entropy, it should be between different distributions, it is weird to have it between embedding and labels?

JoanFM:
Check if labels are needed

finetuner/tuner/pytorch/losses.py
```diff
@@ -174,3 +174,46 @@ def get_default_miner(
...
+    def __init__(self, temperature: float = 0.5) -> None:
+        """Initialize the loss instance.
+
+        :param temerature: The temperature parameter
+        """
+        super().__init__()
+
+        self.temperature = temperature
+
+    def forward(self, embeddings: torch.Tensor, labels: torch.Tensor) -> torch.Tensor:
```
JoanFM, line 196: It may even be much easier to reason about this loss if the interface shows `left_embeddings` and `right_embeddings` so that u know that by `zipping` them u can find the positive correspondances

JoanFM:
Would this work?

```pyhon
class NTXentLoss(nn.Module):
    def __init__(self, temperature: float = 0.5) -> None:
        super().__init__()
        self.temperature = temperature
        self.log_softmax = torch.nn.LogSoftmax(dim=1)


    def forward(self, embeddings_left, embeddings_right):
       # maybe make sure to normalize embeddings
        all_dot_products = embeddings_left.matmul(embeddings_right.T)
        return torch.mean(torch.diagonal(-self.log_softmax(all_dot_products / self.temperature), 0))
```

bwanglzu:
change the default temperature to `0.1`? simclr and v2 both indicates temperature below 1 is good, but 0.1 works best:
> the optimal temperature in {0.1, 0.5, 1.0} is 0.5 and seems consistent regardless of the batch sizes. However, the performance with τ = 0.1 improves as batch size increases, which may suggest a small shift of optimal temperature towards 0.1.

finetuner/tuner/pytorch/losses.py
```diff
@@ -174,3 +174,45 @@ def get_default_miner(
...
+        labels1, labels2 = labels.unsqueeze(1), labels.unsqueeze(0)
+
+        pos_samples = (labels1 == labels2).to(sim.dtype) - diag
+
+        if not (pos_samples.sum(axis=1) == 1).all().item():
+            raise ValueError('There need to be two views of each label in the batch.')
+
+        self_mask = torch.ones_like(sim, requires_grad=False) - diag
+        upper = torch.sum(sim * pos_samples, dim=1)
+        lower = torch.log(torch.sum(self_mask * torch.exp(sim), dim=1))
```
bwanglzu, line 216: is it correct? I think `upper` should be `torch.exp(torch.sum(sim * pos_samples), dim=1)` and `lower` should be without `torch.log` and move it outside.

If I understand correctly, `upper` and `lower` means `numerator ` and `denominator` respectively

bwanglzu (commented):
some comments regards to the loss function

finetuner/tuner/paddle/losses.py
```diff
@@ -187,3 +187,48 @@ def get_default_miner(
...
+
+        sim = (1 - get_distance(embeddings, 'cosine')) / self.temperature
+        diag = paddle.eye(sim.shape[0], dtype=sim.dtype)
+        diag = paddle.to_tensor(diag, place=sim.place)
+        labels1, labels2 = labels.unsqueeze(1), labels.unsqueeze(0)
+
+        pos_samples = paddle.cast(labels1 == labels2, sim.dtype) - diag
+
+        if not (pos_samples.sum(axis=1) == 1).all().item():
+            raise ValueError('There need to be two views of each label in the batch.')
```
bwanglzu, line 228: i think it dependents on the `views_per_instance`? this need to change from:
> There need to be two views of each label in the batch.
to
>There need to be at least two views of each label in the batch.

bwanglzu (changes requested):
in general looks good, left some comments

gmastrapas (changes requested):
just a typo

finetuner/tuner/paddle/losses.py
tadejsv: Right now this loss only works properly with exactly two views (that's how it was designed originally). We can extend it to more than two views (PML does it), but I think the original implementation is good for now

bwanglzu (approved):
LGTM!

gmastrapas (approved):

PR #331, opened 2022-01-10
feat: add default preprocess fn for ssl

bwanglzu:
add a default vision preprocessor, used for self-supervised learning.


finetuner/tuner/augmentation.py
```diff
@@ -0,0 +1,45 @@
...
+        [
+            A.HorizontalFlip(p=0.5),
+            A.ColorJitter(p=1, brightness=0, contrast=0, saturation=0, hue=0),
+            A.RandomResizedCrop(width=width, height=height, p=1),
+            A.GaussianBlur(p=1),
+            A.GridDropout(p=0.5),
+        ]
+    )
+    doc.blob = transform(image=doc.blob)['image']
+    return doc
```
gmastrapas, line 45: Shouldn't this return a NumPy array? This is the interface of `preprocess_fn` right?

gmastrapas (approved):

finetuner/tuner/augmentation.py
bwanglzu: good catch!

tadejsv (approved):
Seems good

PR #338, opened 2022-01-12
feat: use docarray package and remove labeler

tadejsv:
This replaces `DocumentArray` from `jina` with the one from `docarray`.

Also, the **labeler** and `DocumentArrayMemmap` ***are removed***.

Main changes, apart from switching imports:
- simplify `DocumentSequence` into simply `DocumentArray`
- replace a lot of `blob` attributes with `tensor`


setup.py
```diff
@@ -39,7 +39,7 @@
     long_description_content_type='text/markdown',
     zip_safe=False,
     setup_requires=['setuptools>=18.0', 'wheel'],
-    install_requires=['jina>=2.4.9', 'matplotlib'],
+    install_requires=['jina>=2.6.5.dev60', 'docarray>=0.4.2'],
```
bwanglzu, line 42: why `jina` is required?

bwanglzu (changes requested):
shall we also remove `https://github.com/jina-ai/jina/archive/refs/heads/master.zip` from `requirements-cicd(gpu)`?

setup.py
tadejsv: Ah good point, forgot to remove

bwanglzu (approved):
LGTM!

PR #340, opened 2022-01-13
fix: add normalization to preprocessor

bwanglzu:
a set of minor updates based on experiments:

1. change `num_layers` of `projection_head` to 2. based on simclrv2 paper.
2. added another `batchnorm1d` at the end of default `projection_head`.
3. add `phrase` option in `vision_preprocessor`, to apply augmentation at evaluation time.

### DIFF
--- a/finetuner/tuner/augmentation.py
+++ b/finetuner/tuner/augmentation.py
@@ -7,7 +7,7 @@ def vision_preprocessor(
     width: int = 224,
     default_channel_axis: int = -1,
     target_channel_axis: int = 0,
-    phrase: str = 'train',
+    phase: str = 'train',
 ):
     """Randomly augmentation a Document with `blob` field.
     The method applies flipping, color jitter, cropping, gaussian blur and random rectangle erase
@@ -17,13 +17,13 @@ def vision_preprocessor(
     :param width: image width.
     :param default_channel_axis: The color channel of the input image, by default -1, the expected input is H, W, C.
     :param target_channel_axis: The color channel of the output image, by default 0, the expected output is C, H, W.
-    :param phrase: phrase of experiment, either `train` or `validation`. At `validation` phrase, will not apply
+    :param phase: phase of experiment, either `train` or `validation`. At `validation` phase, will not apply
       random transformation.
     """
 
     def preprocess_fn(doc):
         return _vision_preprocessor(
-            doc, height, width, default_channel_axis, target_channel_axis, phrase
+            doc, height, width, default_channel_axis, target_channel_axis, phase
         )
 
     return preprocess_fn
@@ -51,7 +51,7 @@ def _set_image_blob_normalization(
     .. warning::
         Please do NOT generalize this function to gray scale, black/white image, it does not make any sense for
         non RGB image. if you look at their MNIST examples, the mean and stddev are 1-dimensional
-        (since the inputs are greyscale-- no RGB channels).
+        (since the inputs are greyscale there is no RGB channels).
 
 
     """
@@ -71,7 +71,7 @@ def _vision_preprocessor(
     width: int = 224,
     default_channel_axis: int = -1,
     target_channel_axis: int = 0,
-    phrase: str = 'train',
+    phase: str = 'train',
 ):
     """Randomly augmentation a Document with `blob` field.
     The method applies flipping, color jitter, cropping, gaussian blur and random rectangle erase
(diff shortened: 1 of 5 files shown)


finetuner/tuner/augmentation.py
```diff
@@ -55,6 +55,8 @@ def _vision_preprocessor(
             blob = doc.blob
         else:
             raise AttributeError('Can not load `blob` field from the given document.')
```
gvondulong, line 57: I just saw this in the previous PR. Can you maybe show the content of the blob field to the user or somehow show the user which case (url or blob) failed?

finetuner/tuner/augmentation.py
```diff
@@ -16,22 +17,61 @@ def vision_preprocessor(
...
+
+    :param channel_axis: the axis id of the color channel, ``-1`` indicates the color channel info at the last axis
+    :param img_mean: the mean of all images
+    :param img_std: the standard deviation of all images
+    :return: itself after processed
+
+    .. warning::
+        Please do NOT generalize this function to gray scale, black/white image, it does not make any sense for
+        non RGB image. if you look at their MNIST examples, the mean and stddev are 1-dimensional
+        (since the inputs are greyscale-- no RGB channels).
```
gvondulong, line 54: since the inputs are greyscale `there is` no RGB channels

finetuner/tuner/augmentation.py
```diff
@@ -16,22 +17,61 @@ def vision_preprocessor(
...
+
+
+    """
+    blob = (blob / 255.0).astype(np.float32)
```
gvondulong, line 58: Will this always be float32?

finetuner/tuner/augmentation.py
```diff
@@ -16,22 +17,61 @@ def vision_preprocessor(
     :param width: image width.
     :param default_channel_axis: The color channel of the input image, by default -1, the expected input is H, W, C.
     :param target_channel_axis: The color channel of the output image, by default 0, the expected output is C, H, W.
+    :param phrase: phrase of experiment, either `train` or `validation`. At `validation` phrase, will not apply
```
gmastrapas, line 20: do you mean `phase`?

finetuner/tuner/augmentation.py
```diff
@@ -16,22 +17,61 @@ def vision_preprocessor(
...
+        non RGB image. if you look at their MNIST examples, the mean and stddev are 1-dimensional
+        (since the inputs are greyscale-- no RGB channels).
+
+
+    """
+    blob = (blob / 255.0).astype(np.float32)
+    blob = np.moveaxis(blob, channel_axis, 0)
+    mean = np.asarray(img_mean, dtype=np.float32)
+    std = np.asarray(img_std, dtype=np.float32)
+    blob = (blob - mean[:, None, None]) / std[:, None, None]
```
gvondulong, line 62: Imho, the normalization should be done with a pixel value mean that is computed of the whole dataset. 
If one image is super dark and another one is very bright, both will end up to show average lighting conditions, when normalizing per-image.

gmastrapas (changes requested):

gvondulong (commented):
Looks good. Just one point about types and one on normalization.

finetuner/tuner/augmentation.py
bwanglzu: good catch, this is something i constantly spell incorrectly :)

gmastrapas: both phase and phrase are greek, so :P

finetuner/tuner/augmentation.py
bwanglzu: could be, but it's tricky to get global mean value in a dataset. This piece of code is actually copied from jina core (it's in the image mixin). @gvondulong

tadejsv: Yeah this is the standard imagenet normalization - it's used in nearly all CV applications (unless your domain is something completely different, like tissue cells)

finetuner/tuner/augmentation.py
```diff
@@ -16,22 +17,61 @@ def vision_preprocessor(
...
     def preprocess_fn(doc):
         return _vision_preprocessor(
-            doc, height, width, default_channel_axis, target_channel_axis
+            doc, height, width, default_channel_axis, target_channel_axis, phase
         )
 
     return preprocess_fn
 
 
+def _set_image_blob_normalization(
```
tadejsv, line 32: This can actually also be done with albumentations (for example [`Normalize`](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.Normalize) and [`ToFloat`](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.ToFloat)). Wouldn't it be better to use that?

gvondulong (approved):

gmastrapas (approved):

tadejsv (approved):

PR #342, opened 2022-01-18
fix: do not normalize float images

bwanglzu:
while doing experimentation, found out the `Normalize` function in `albumentations` produce different result than jina `normlaize` function. Unified all with the default jina version.


finetuner/tuner/augmentation.py
```diff
@@ -63,18 +73,13 @@ def _vision_preprocessor(
...
+        blob = doc.blob
     if blob.dtype == np.float64:
         blob = np.float32(blob)
     if default_channel_axis not in [-1, 2]:
         blob = np.moveaxis(blob, default_channel_axis, -1)
-    if blob.shape[-1] == 3:  # apply to RGB images.
-        transform = A.Compose(
-            [
-                A.Normalize(),
-                A.ToFloat(),
```
tadejsv, line 74: I think the problem is that `ToFloat` here is not necessary. Can you debug a bit more?

finetuner/tuner/augmentation.py
```diff
@@ -7,6 +7,7 @@ def vision_preprocessor(
     width: int = 224,
     default_channel_axis: int = -1,
     target_channel_axis: int = 0,
+    normalize: bool = False,
     phase: str = 'train',
 ):
     """Randomly augmentation a Document with `blob` field.
```
gmastrapas, line 13: augment

finetuner/tuner/augmentation.py
```diff
@@ -35,6 +43,7 @@ def _vision_preprocessor(
...
```
gmastrapas, line 49: augment

gmastrapas (approved):

gvondulong (commented):
lgtm!

finetuner/tuner/augmentation.py
bwanglzu: we stick with docarray? this gives me more confident that we're doing the same thing as the common preprocess chain, such as `doc.convert_uri_to_tensor().set_tensor_normlzation().set_channel_axis()..`

tadejsv (approved):
LGTM

PR #344, opened 2022-01-19
feat: custom tag key

florian-hoenicke:


gmastrapas (changes requested):
PyTorch and PaddlePaddle tuners should also accept this argument, I think the Evaluator too

florian-hoenicke:
> PyTorch and PaddlePaddle tuners should also accept this argument, I think the Evaluator too

Thanks. I will add it.

numb3r3:
BTW, a preprocess function could also be applied to do this job. 

```
def preprocess(doc: Document)
    doc.tags['finetuner_tag'] = doc.tags['custom_tag']
    return doc
```

bwanglzu:
@florian-hoenicke let's develop new features in `finetuner.fit`

florian-hoenicke:
Okay, I'll move the pr

PR #351, opened 2022-01-25
fix: shuffle batches

florian-hoenicke:
Before:
image

After:
image

### DIFF
--- a/finetuner/tuner/dataset/samplers.py
+++ b/finetuner/tuner/dataset/samplers.py
@@ -1,3 +1,4 @@
+import random
 from collections import defaultdict
 from copy import deepcopy
 from math import ceil
@@ -150,6 +151,7 @@ class ClassSampler(BaseSampler):
             batches.append(batch)
 
         self._batches = batches
+        random.shuffle(self._batches)
 
 
 class SessionSampler(BaseSampler):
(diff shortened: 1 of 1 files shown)


finetuner/tuner/dataset/samplers.py
```diff
@@ -150,6 +151,7 @@ def _prepare_batches(self):
             batches.append(batch)
 
         self._batches = batches
+        random.shuffle(self._batches)
```
JoanFM, line 154: there should be some option right? sometimes to debug is nice to disable?

finetuner/tuner/dataset/samplers.py
```diff
@@ -1,3 +1,4 @@
+import random
```
tadejsv, line 1: `shuffle` is already imported

tadejsv (changes requested):

finetuner/tuner/dataset/samplers.py
florian-hoenicke: thanks

finetuner/tuner/dataset/samplers.py
florian-hoenicke: I don't think so. There will be no advantage of having shuffling disabled

tadejsv (approved):

PR #362, opened 2022-02-04
docs: change ResNet18 to ResNet50 in README exmaple

guenthermi:
Since tf.keras.applications does not provide ResNet18, I changed the readme example to use ResNet50, which is supported by all three frameworks. Moreover, I derived the input_size parameter from the input data since the input shape is different for keras.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ pip install finetuner
             doc.load_uri_to_image_tensor(224, 224)
             .set_image_tensor_normalization()
             .set_image_tensor_channel_axis(-1, 0)
-        )  # No need for changing channel axes line if you are using tf/keras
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -94,8 +94,7 @@ pip install finetuner
         device='cuda',
         batch_size=128,
         to_embedding_model=True,
-        input_size=(3, 224, 224),
-        layer_name='adaptiveavgpool2d_67', # layer before fc as feature extractor
+        input_size=data.tensors.shape[1:],
```
gmastrapas, line 97: I think this should still be explicit

guenthermi: Ok, I changed it with the last commit

gmastrapas (approved):
Looks good 🚀

bwanglzu (approved):
LGTM!

PR #378, opened 2022-02-25
fix: fix the docarray version to pass unit tests

makram93:
downgrade docarray due to error in `match` function.


bwanglzu (approved):
LGTM!

PR #387, opened 2022-04-25
feat: create run endpoints

jupyterjazz:
closes https://github.com/jina-ai/finetuner.fit/issues/182 and https://github.com/jina-ai/finetuner.fit/issues/184


tests/conftest.py
```diff
@@ -9,5 +12,12 @@ def handle_request_mocker(self, **kwargs):
...
 
     mocker.patch.object(Client, 'handle_request', handle_request_mocker)
-    client = Client(user_id=1)
+    client = Client(user_id=USER_ID_FOR_TESTING)
     return client
+
+
+@pytest.fixture
+def run_config(path=RUN_CONFIG_PATH):
+    with open(path) as f:
```
bwanglzu, line 21: minor thing, i think `RUN_CONFIG_PATH` is not needed to be parsed as an argument since you define it as a constant

tests/integration/test_runs.py
```diff
@@ -0,0 +1,87 @@
...
+        == run_config
+    )
+    # delete the first run
+    client.delete_run(experiment_name=experiment_name, run_name=first_run)
+    response = client.list_runs(experiment_name=experiment_name)
+    exp_runs = response[0].json()
+    assert len(exp_runs) == 1
+    assert exp_runs[0]['name'] == second_run
+    # delete all existing runs
+    client.delete_runs(experiment_name=experiment_name)
```
bwanglzu, line 47: after delete run, can we somehow assert the `STATUS` in AWS batch to be inactive or failed? (make sure it is canceled)

bwanglzu (changes requested):
in general looks good to me, minor comments. We'll need to refactor the interface to allow user call `finetuner.fit` rather than provide a json. Let's do it after we do the hubble client integration

bwanglzu (approved):
LGTM! let's remove user id and add hubble client in the next iteration!

PR #388, opened 2022-04-27
feat: hubble integration

jupyterjazz:
Closes https://github.com/jina-ai/finetuner.fit/issues/185

Main things added:
- Hubble authentication
- Push train (and evaluation) data to Hubble
- Pull finetuned model from Hubble
- Integration tests use staging env


finetuner/client/__init__.py
```diff
@@ -0,0 +1,3 @@
+import hubble
+
+hubble.login()
```
bwanglzu, line 3: maybe i's better to ask user to decide login rather than call the login automatically and open the browser?

finetuner/client/base.py
```diff
@@ -1,23 +1,39 @@
...
-        self._user_id = user_id
-        self._session = requests.session()
+    def __init__(self):
         self._base_url = Path(os.environ.get(HOST))
-        self._session.headers.update({'Accept-Charset': 'utf-8'})
+        self._session = self._get_client_session()
+        self._hubble_client = hubble.Client(max_retries=None, timeout=10, jsonify=True)
+        self._hubble_user_id = json.loads(self._hubble_client.get_user_info())[DATA][
+            HUBBLE_USER_ID
+        ]
```
bwanglzu, line 29: what will happen if we get a 404 here? json load `[DATA][HUBBLE_USER_ID]` will break, right?

finetuner/client/client.py
```diff
@@ -56,39 +59,63 @@ def delete_experiment(self, name: str):
...
+        return self.handle_request(url=url, method=DELETE)
 
     def create_run(
-        self, experiment_name: str, run_name: str, config: Dict[str, Any], **kwargs
+        self,
+        model: str,
+        train_data: Union[DocumentArray, str],
+        experiment_name: str,
+        run_name: str,
+        **kwargs,
```
bwanglzu, line 78: all other parameters will be fold into kwargs?

finetuner/client/client.py
```diff
@@ -175,4 +194,47 @@ def get_run_status(self, experiment_name: str, run_name: str):
...
+                [self._hubble_user_id, experiment_name, run_name, TRAIN_DATA]
+            )
+            train_data.push(name=da_name)
+            train_data = da_name
+        if isinstance(eval_data, DocumentArray):
+            da_name = '-'.join(
+                [self._hubble_user_id, experiment_name, run_name, EVAL_DATA]
+            )
+            eval_data.push(name=da_name)
+            eval_data = da_name
```
bwanglzu, line 225: let's first assume they're always `DocumentArray`, not `str`, otherwise `str` is not handled here

bwanglzu: and `eval_data` is `Optional`, make sure it is checked `if eval_data and isinstance...`

requirements-dev.txt
```diff
@@ -5,3 +5,4 @@ isort==5.10.1
 pytest==7.0.0
 pytest-cov==3.0.0
 pytest-mock==3.7.0
+google-api-python-client==2.46.0
```
bwanglzu, line 8: why do we need this?

requirements.txt
```diff
@@ -1,3 +1,7 @@
 requests==2.27.1
 path==16.4.0
-python-dotenv==0.19.2
\ No newline at end of file
+python-dotenv==0.19.2
+jina-hubble-sdk==0.1.3.dev8
+pytest==7.0.0
```
bwanglzu, line 5: make sure only "mandantory" libraries in `requirements.txt`, others move to `requirements-dev` or `requirements-extra.txt`

bwanglzu (changes requested):
great work! left some comments

guenthermi (changes requested):
Good work, added some comments

finetuner/client/client.py
```diff
@@ -56,39 +59,63 @@ def delete_experiment(self, name: str):
...
+        :param train_data: Either a `DocumentArray` for training data or a
+                           name of the `DocumentArray` that is pushed on Hubble.
         :param experiment_name: The name of the experiment.
         :param run_name: The name of the run.
-        :param config: Configuration for the run.
         :return: `requests.Response` object.
         """
+        train_data, kwargs[EVAL_DATA] = self._push_data_to_hubble(
+            train_data=train_data,
+            eval_data=kwargs.get(EVAL_DATA),
```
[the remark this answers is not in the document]
jupyterjazz: `.get()` returns `None` if the key is not found. I'm not sure how this is supposed to fail..

requirements-dev.txt
jupyterjazz: Otherwise I have this error:
error

finetuner/client/client.py
jupyterjazz: Sure but it works as expected - passing [these tests](https://github.com/jina-ai/finetuner/blob/2b8625ac62c0db02b86f49148e51ddbfb17b8ba3/tests/unit/test_hubble.py#L7) (One thing that's missing for strings is checking if the name is valid) - so maybe better to leave it as it is for now and refactor it later?

Added `if eval_data ...`

requirements-dev.txt
jupyterjazz: Needed during `da.push()`

finetuner/client/client.py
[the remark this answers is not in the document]
guenthermi: oh, you are right

requirements-dev.txt
bwanglzu: docarray did not install it by default?

tests/integration/conftest.py
```diff
@@ -0,0 +1,54 @@
...
+from tests.constants import FINETUNER_LABEL
+
+import finetuner
+from finetuner.client.client import Client
+
+
+@pytest.fixture()
+def get_image_data(data_path='tests/resources/image_data/'):
+    left_da = DocumentArray.from_files(os.path.join(data_path, 'left/*.jpg'))
+    right_da = DocumentArray.from_files(os.path.join(data_path, 'right/*.jpg'))
```
bwanglzu, line 15: no please remove these images from resources, it makes repo too big to clone. You can generate some random images using numpy `np.random.rand(3, 224, 224)`

bwanglzu (changes requested):
left some comments

requirements-dev.txt
gmastrapas: should it be in `dev` though?

finetuner/client/base.py
```diff
@@ -1,31 +1,53 @@
...
+        api_token = TOKEN_PREFIX + str(hubble.Auth.get_auth_token())
+        session.headers.update({CHARSET: UTF_8, AUTHORIZATION: api_token})
+        return session
 
     def handle_request(
         self,
         url: str,
         method: str,
         params: Optional[dict] = None,
         json: Optional[dict] = None,
```
gmastrapas, line 49: can we rename this argument? it shadows the built-in library

finetuner/client/base.py
```diff
@@ -1,31 +1,53 @@
...
-    ):
+    ) -> requests.Response:
         """The basis request handler.
```
gmastrapas, line 51: base

finetuner/client/base.py
```diff
@@ -1,31 +1,53 @@
...
+        return session
 
     def handle_request(
         self,
         url: str,
         method: str,
         params: Optional[dict] = None,
         json: Optional[dict] = None,
-    ):
+    ) -> requests.Response:
```
gmastrapas, line 50: I think we should return dicts. No reason for the user to have to deal with custom `requests` objects

finetuner/client/base.py
```diff
@@ -1,31 +1,53 @@
...
+        return hubble_user_id
+
+    @staticmethod
+    def _get_client_session() -> requests.Session:
+        session = requests.Session()
+        api_token = TOKEN_PREFIX + str(hubble.Auth.get_auth_token())
+        session.headers.update({CHARSET: UTF_8, AUTHORIZATION: api_token})
+        return session
 
     def handle_request(
```
gmastrapas, line 44: I would make `handle_request` private

finetuner/client/client.py
```diff
@@ -27,69 +34,90 @@ def create_experiment(self, name: str, **kwargs):
...
+        model: str,
+        train_data: str,
+        **kwargs,
+    ) -> Dict[str, Any]:
+        # Not sure what is the correct way to construct the whole config.
+        # Maybe we can create a separate ticket for this.
+        config = {}
+        config[MODEL] = model
+        config[DATA] = {TRAIN_DATA: train_data, EVAL_DATA: kwargs.get(EVAL_DATA)}
+        return config
```
gmastrapas, line 118: what about the rest of the kwargs?

gmastrapas (commented):
Left some comments

finetuner/client/base.py
jupyterjazz: Agree, they should at least have option to receive dicts. But the PR is too big now and maybe we should address it next time?

finetuner/client/client.py
jupyterjazz: We'll revisit it later

requirements-dev.txt
jupyterjazz: I'll ask about it on slack

jupyterjazz: I asked for help on the docarray channel. So maybe we can ignore it for now and I'll revisit that later.

jupyterjazz: @gmastrapas 
> should it be in dev though?'

If it's needed during `da.push()`, we should have it in `requirements.txt`, good point.

gmastrapas (approved):
Looks good!

PR #394, opened 2022-05-17
feat: implement basic error handling

jupyterjazz:
Implements basic error handling.
We can come up with more sophisticated error classes/types after.

### DIFF
--- a/finetuner/client/base.py
+++ b/finetuner/client/base.py
@@ -46,9 +46,7 @@ class _BaseClient:
 
     @staticmethod
     def _handle_error_request(response: requests.Response):
-        message = response.reason
-        code = response.status_code
-        raise BaseError(response=response.json(), message=message, code=code)
+        raise BaseError(message=response.reason, code=response.status_code)
 
     def _handle_request(
         self,
--- a/finetuner/client/base.py
+++ b/finetuner/client/base.py
@@ -6,6 +6,7 @@ import requests
 from path import Path
 
 import hubble
+from finetuner.client.exception import BaseError
 from finetuner.constants import (
     AUTHORIZATION,
     CHARSET,
@@ -43,6 +44,12 @@ class _BaseClient:
         session.headers.update({CHARSET: UTF_8, AUTHORIZATION: api_token})
         return session
 
+    @staticmethod
+    def _handle_error_request(response: requests.Response):
+        message = response.reason
+        code = response.status_code
+        raise BaseError(response=response.json(), message=message, code=code)
+
     def _handle_request(
         self,
         url: str,
@@ -60,5 +67,8 @@ class _BaseClient:
         """
         response = self._session.request(
             url=url, method=method, json=json_data, params=params, verify=False
-        ).json()
-        return response
+        )
+        if not response.ok:
+            self._handle_error_request(response)
+
+        return response.json()
--- a/finetuner/client/base.py
+++ b/finetuner/client/base.py
@@ -64,7 +64,9 @@ class _BaseClient:
         )
         if not response.ok:
             raise FinetunerServerError(
-                message=response.reason, code=response.status_code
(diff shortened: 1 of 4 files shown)


finetuner/client/exception.py
```diff
@@ -0,0 +1,24 @@
+from typing import Optional
+
+
+class BaseError(Exception):
+    response = None
+    code = -1
+    message = "An unknown error occurred"
```
gmastrapas, line 7: single quotes

finetuner/client/exception.py
```diff
@@ -0,0 +1,24 @@
+from typing import Optional
+
+
+class BaseError(Exception):
```
gmastrapas, line 4: I would name this `FinetunerServerError` for now

finetuner/client/exception.py
```diff
@@ -0,0 +1,24 @@
...
+    response = None
+    code = -1
+    message = "An unknown error occurred"
+
+    def __init__(
+        self,
+        response: dict,
+        message: Optional[str] = None,
```
gmastrapas, line 12: setting default None values makes the code above more complex. Just set the class attribute values as default here:
```
class BaseError(Exception):

def __init__(
        self,
        response: dict,
        message: str = "An unknown error occurred",
        code: int = -1,
```

finetuner/client/base.py
```diff
@@ -60,5 +67,8 @@ def _handle_request(
         """
         response = self._session.request(
             url=url, method=method, json=json_data, params=params, verify=False
-        ).json()
-        return response
+        )
+        if not response.ok:
+            self._handle_error_request(response)
```
gmastrapas, line 72: Just do here:

```
raise BaseError(response=response.json(), message=response.reason, code=response.status_code)
```
no need to have an additional function

finetuner/run.py
```diff
@@ -53,10 +53,7 @@ def save_model(self, path: str = FINETUNED_MODELS_DIR):
         :returns: A list of str object(s) that indicate the download path.
         """
         if self.status() != FINISHED:
-            # tell users that the run hasn't finished yet, thus they can't download
-            # the model.
-            # will implement a proper error-handling in a separate pr.
-            raise ValueError
+            raise Exception("The run needs to be finished in order to save the model.")
```
gmastrapas, line 56: single quotes

gmastrapas (changes requested):

jupyterjazz:
Looks like this
servererror

gmastrapas:
> Looks like this servererror

Hmm I would say let's do it like this:
```
finetuner.client.exception.FinetunerServerError: INTERNAL SERVER ERROR (500) - details
```

jupyterjazz:
@gmastrapas 
> Hmm I would say let's do it like this:
> 
> ```
> finetuner.client.exception.FinetunerServerError: INTERNAL SERVER ERROR (500) - details
> ```

`Internal Server Error` is the `message` (i.e. `details`) here.

If we change the way you suggested, we'll get something like this:
> finetuner.client.exception.FinetunerServerError: INTERNAL SERVER ERROR (500) - Internal Server Error

gmastrapas:
> @gmastrapas
> 
> > Hmm I would say let's do it like this:
> > ```
> > finetuner.client.exception.FinetunerServerError: INTERNAL SERVER ERROR (500) - details
> > ```
> 
> `Internal Server Error` is the `message` (i.e. `details`) here.
> 
> If we change the way you suggested, we'll get something like this:
> 
> > finetuner.client.exception.FinetunerServerError: INTERNAL SERVER ERROR (500) - Internal Server Error

Inside the body there must be a `detail` field

jupyterjazz:
@gmastrapas good point!

This is the new one
newerror

gmastrapas (approved):
Looks good!

PR #396, opened 2022-05-17
test: improve unit testing

jupyterjazz:
Closes https://github.com/jina-ai/finetuner.fit/issues/227


tests/unit/conftest.py
```diff
@@ -1,10 +1,25 @@
...
+    CREATED_AT,
+    DESCRIPTION,
+    FAILED,
+    FINISHED,
+    NAME,
+    RUN_NAME,
+    STARTED,
+    STATUS,
+)
+from finetuner.finetuner import Finetuner
```
gmastrapas, line 22: why do we need 2 mock objects? isn't mock_client enough for testing?

tests/unit/conftest.py
```diff
@@ -25,5 +40,55 @@ def get_auth_token():
     mocker.patch.object(docarray.DocumentArray, 'push', return_args)
     hubble.login()
     client = FinetunerV1Client()
-    mocker.patch.object(client, 'hubble_user_id', '1')
+    mocker.patch.object(client, 'hubble_user_id', HUBBLE_USER_TEST_ID)
     return client
+
+
+@pytest.fixture
+def test_finetuner(test_client, mocker):
```
gmastrapas, line 48: I would define the mock objects in a separate `mocks` directory. These aren't exactly fixtures they are mocks

jemmyshin (approved):
LGTM!

tests/unit/conftest.py
jupyterjazz: The first mock object in on the client-level and is used to test how we send requests - checks if url is constructed correctly, if data is passed, etc.

The second mock is on the finetuner-level and is used to imitate what client would've returned to test how creating/deleting/listing things work.

tests/unit/conftest.py
jupyterjazz: How do you suggest using mock objects without fixtures? Do we want to create a new mock object for each test? 

> These aren't exactly fixtures they are mocks

I don't think the purpose of fixtures exclude using mock objects i.e. fixtures can be mocks.

tests/unit/conftest.py
gmastrapas: The 2 external services that we use in this package is hubble and finetuner-api. So we need 2 mock objects, the first one mocking the hubble SDK and the second one mocking the API. It is difficult to mock the API so you could mock the `FinetunerV1Client` with a mock class that holds the experiments and the runs internally in-memory. I think it is conceptually simpler and more comprehensive this way. For example see how we mock the scheduler in the api

jupyterjazz:
> Should we fix the tests?

Unit tests are passing locally. I think github secrets need to be updated here..
I'll make a separate PR for integration tests.

gmastrapas (approved):

PR #401, opened 2022-05-23
fix: `download_artifact` throws error

jupyterjazz:
Closes https://github.com/jina-ai/finetuner.fit/issues/232


gmastrapas (commented):
So the problem was that we weren't deserializing?

jupyterjazz:
> So the problem was that we weren't deserializing?

yeap

gmastrapas (approved):
Looks good!

jemmyshin (approved):

violenil (approved):
LGTM!

PR #411, opened 2022-05-26
feat: expose arguments `cpu`, `num_workers` and `queue` in `finetuner.fit`

### LINKED ISSUE #409: Enable GPU runs
We should allow the use of GPUs and the option to specify a number of CPUs in the client. API already supports this

gmastrapas:
Closes #409


bwanglzu (changes requested):
can we remove `queue`?

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Expose arguments `cpu`, `num_workers` and `queue` in `finetuner.fit`. ([#411](https://github.com/jina-ai/finetuner/pull/411))
```
jupyterjazz, line 13: Can you update this line as well?

tests/unit/test_client.py
```diff
@@ -61,6 +61,9 @@ def test_create_run(client_mocker, experiment_name='exp', run_name='run'):
         experiment_name=experiment_name,
         run_name=run_name,
         run_config=config,
+        device='cpu',
+        cpus=1,
+        gpus=1,
```
jupyterjazz, line 66: Can you also check that they were sent correctly? i.e. `assert sent_request['json_data'][GPUS] == gpus`

finetuner/experiment.py
```diff
@@ -152,10 +155,17 @@ def create_run(
...
+        cpu = kwargs.get(CPU, True)
+        num_workers = kwargs.get(NUM_WORKERS, 4)
+
         run_info = self._client.create_run(
             run_name=run_name,
             experiment_name=self._name,
             run_config=config,
+            device='cpu' if cpu else 'gpu',
+            cpus=num_workers,
+            gpus=1,
```
jupyterjazz, line 168: Why should we bother having `gpus` here if it's always 1?

bwanglzu (commented):
why did we remove `wandb` option? it's not working?

finetuner/experiment.py
gmastrapas: At some point it will be used

gmastrapas:
> why did we remove `wandb` option? it's not working?

I'd rather not prompt users to give us their wandb tokens. We need to find a better way for this

bwanglzu (approved):
LGTM! in the finetuner runner, we need to parse these two parameters as well.

jupyterjazz (approved):
:rocket:

PR #415, opened 2022-06-01
feat: support callbacks

jupyterjazz:
closes https://github.com/jina-ai/finetuner.fit/issues/238

```python
import finetuner
from finetuner.callbacks import TrainingCheckpoint, BestModelCheckpoint
training_chckpt = TrainingCheckpoint(last_k_epochs=3)

run = finetuner.fit(..., callbacks=[training_chckpt, BestModelCheckpoint()]
```
Connected with https://github.com/jina-ai/finetuner.fit/pull/241


jupyterjazz:
@bwanglzu @gmastrapas 
What if we created callback class templates, so that users would do something like this

```python
import finetuner
from finetuner.callbacks import TrainingCheckpoint, BestModelCheckpoint
training_chckpt = TrainingCheckpoint(last_k_epochs=3)

run = finetuner.fit(..., callbacks=[training_chckpt, BestModelCheckpoint()]
```

gmastrapas:
> @bwanglzu @gmastrapas What if we created callback class templates, so that users would do something like this
> 
> ```python
> import finetuner
> from finetuner.callbacks import TrainingCheckpoint, BestModelCheckpoint
> training_chckpt = TrainingCheckpoint(last_k_epochs=3)
> 
> run = finetuner.fit(..., callbacks=[training_chckpt, BestModelCheckpoint()]
> ```

Ι like it

bwanglzu:
looks good

finetuner/callbacks.py
```diff
@@ -0,0 +1,131 @@
+from typing import Optional
+
+
+class BestModelCheckpoint:
```
gmastrapas, line 4: I would use `dataclass` for the templates. It will provide validation features

jupyterjazz: haven't really used dataclasses before..

finetuner/callbacks.py
```diff
@@ -0,0 +1,160 @@
...
+    :param mode: one of {'auto', 'min', 'max'}. The decision to overwrite the
+        currently saved model is made based on either the maximization or the
+        minimization of the monitored quantity.
+        For an evaluation metric, this should be `max`, for `val_loss` this should
+        be `min`, etc. In `auto` mode, the mode is set to `min` if `monitor='loss'`
+        or `monitor='val_loss'` and to `max` otherwise.
+    :param verbose: Whether to log notifications when a checkpoint is saved.
+    """
+
+    name: str = 'BestModelCheckpoint'
```
gmastrapas, line 20: you don't have to specify `name` on any callback. You can always get like this `callback.__class__.__name__`

finetuner/callbacks.py
```diff
@@ -0,0 +1,160 @@
...
+    save_dir: str = 'best_model_chckpt'
+    monitor: str = 'val_loss'
+    mode: str = 'auto'
+    verbose: bool = False
+
+    def get_options(self) -> dict:
```
gmastrapas, line 26: you don't have to specify this method every time. You can get the attrs of a dataclass using `fields` https://stackoverflow.com/questions/51946571/how-can-i-get-python-3-7-new-dataclass-field-types

gmastrapas (commented):
Nice 🚀 Just a final thing, in the callback docstrings add also some small descriptions for each one apart from the argument reference

finetuner/callbacks.py
```diff
@@ -0,0 +1,138 @@
...
+        period.
+    :param cloudwatch: Boolean value, if sync logs to aws cloud watch.
+    """
+
+    batch_period: int = 100
+    cloudwatch: bool = False
+
+
+@dataclass
+class ProgressBarCallback:
```
gmastrapas, line 101: Sorry I missed that, this callback should not be exposed to the user

jupyterjazz: makes sense

gmastrapas (approved):
Looks good 🚀 🚀 🚀

bwanglzu (approved):
LGTM!

PR #419, opened 2022-06-02
refactor: explain list models

bwanglzu:
A bit explanation on the support backbone model, attached 2 new models in the runner from @violenil 's transformer PR.

it is needed when writing documentation on the step-by-step guide.

4E947778-F22E-40AC-8F68-B0451650E0E7

### DIFF
--- a/finetuner/__init__.py
+++ b/finetuner/__init__.py
@@ -34,11 +34,32 @@ def login():
 def list_models():
     """List available models for training."""
     return [
-        'resnet50',
-        'resnet152',
-        'efficientnet_b0',
-        'efficientnet_b4',
-        'openai/clip-vit-base-patch32',
+        {
+            'resnet50': '''This model was pre-trained on ImageNet dataset, it is recommended to use the model when you want to fine-tune your model on image-to-image search task. The output feature vector dimensionality is 2048.'''
+        },
+        {
+            'resnet152': '''This model was pre-trained on ImageNet dataset, it is recommended to use the model when you want to fine-tune your model on image-to-image search task. The output feature vector dimensionality is 2048.'''
+        },
+        {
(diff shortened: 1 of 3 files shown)


bwanglzu:
@jupyterjazz i'm not a big fan of loading extra resources files into the repo. the thing here is we need to always align the models here inside runner, this could be potentially an issue.

finetuner/__init__.py
```diff
@@ -34,11 +34,55 @@ def login():
...
+            'task': 'text-to-text',
+            'output_dim': 768,
+            'details': 'Pretrained on BookCorpus and English Wikipedia.',
+        },
+        {
+            'model': 'sentence-transformers/msmarco-distilbert-base-v3',
+            'task': 'text-to-text',
+            'output_dim': 768,
+            'details': '''Pretrained on BookCorpus and English Wikipedia, has been
+            fine-tuned on the msmarco dataset from sentence-transformers.
```
guenthermi, line 83:
```suggestion
            fine-tuned on the MS MARCO dataset from sentence-transformers.
```

guenthermi (approved):
LGTM, added only one minor comment. However, I think at some point, we should expose the available models from an endpoint of the API and it should be connected with the validation of the fine-tuning configuration.

finetuner/__init__.py
```diff
@@ -34,11 +34,55 @@ def login():
...
+            'task': 'image-to-image',
+            'output_dim': 1280,
+            'details': 'Pretrained on ImageNet dataset.',
+        },
+        {
+            'model': 'efficientnet_b0',
+            'task': 'image-to-image',
+            'output_dim': 1280,
+            'details': 'Pretrained on ImageNet dataset.',
+        },
```
jupyterjazz, line 60: Same models

Issue #418, opened 2022-06-01
Support `query_data` and `index_data` in tuner


violenil:
Currently there is only support the use of one evaluation `DocumentArray` in `tuner.fit`. However, it is quite customary to use a set of queries (`query_data`) and a set of indexes (`index_data`) as evaluation data for the `Evaluator`, ie. these are two separate `DocumentArray`s. This option should be supported in finetuner.

gmastrapas:
User will be able to use the evaluation callback which supports index and query data. See https://github.com/jina-ai/finetuner/pull/415

PR #422, opened 2022-06-02
feat: support EvaluationCallback

jupyterjazz:
Closes https://github.com/jina-ai/finetuner/issues/418

I'll resolve/add tests soon (already tested internally)


finetuner/experiment.py
```diff
@@ -180,6 +181,82 @@ def create_run(
...
+        self,
+        train_data: Union[DocumentArray, str],
+        eval_data: Union[DocumentArray, str],
+        eval_callback: EvaluationCallback,
+        run_name: str,
+    ):
+        """Upload data to Hubble and returns their names.
+
+        Uploads all data needed for fine-tuning - training data,
+        evaluation data and query/index data for `EvaluationCallback`.
```
violenil, line 194: Is it really possible to use both `eval_data` as well as `query_data` and `index_data` for this callback? I thought the user either provides `eval_data` or the query + index data. I may be wrong.

finetuner/experiment.py
```diff
@@ -180,6 +181,82 @@ def create_run(
...
+                    client=self._client,
+                    data=eval_data,
+                    data_type=EVAL_DATA,
+                    experiment_name=self._name,
+                    run_name=run_name,
+                )
+        if eval_callback:
+            if not eval_callback.query_data:
+                eval_callback.query_data = (
+                    eval_data_name if eval_data_name else train_data_name
```
violenil, line 229: Is the training data automatically used as eval data, if no eval data is provided?

finetuner/experiment.py
jupyterjazz: When only `eval_data` is provided, `query_data` becomes `eval_data` and `index_data` is `None` afaik.

finetuner/experiment.py
jupyterjazz: Yes, however I'm not sure if it makes sense. 
Do you think we should do something like this?
```python
query_data = eval_data if eval_data else None
```

jupyterjazz: Oh, you asked a different question. You're right, that doesn't make sense.

finetuner/callbacks.py
```diff
@@ -117,3 +119,21 @@ class EarlyStopping:
...
+
+    :param query_data: Search data used by the evaluator at the end of each epoch,
+        to evaluate the model. If `None`, evaluation data will be used as query data.
+        If evaluation data doesn't exist, training data will be chosen.
+    :param index_data: Index data or catalog used by the evaluator at the end of
+        each epoch, to evaluate the model.
+    """
+
+    query_data: Optional[Union[DocumentArray, str]] = None
+    index_data: Optional[Union[DocumentArray, str]] = None
```
gmastrapas, line 139: why not add the rest of the args?

finetuner/experiment.py
```diff
@@ -180,6 +181,79 @@ def create_run(
         )
         return run
 
+    def _prepare_data(
+        self,
+        train_data: Union[DocumentArray, str],
+        eval_data: Union[DocumentArray, str],
+        eval_callback: EvaluationCallback,
```
guenthermi, line 188: `eval_data` and `eval_callback` are `Optional[..]` arguments.

finetuner/experiment.py
```diff
@@ -180,6 +181,79 @@ def create_run(
...
+        Checks not to upload same dataset twice.
+
+        :param train_data: Either a `DocumentArray` for training data or a
+            name of the `DocumentArray` that is pushed on Hubble.
+        :param train_data: Either a `DocumentArray` for evaluation data or a
+            name of the `DocumentArray` that is pushed on Hubble.
+        :param eval_callback: Evaluation callback that contains query and index data
+            and gets modified in-place.
+        :param run_name: Name of the run.
+        :return: Name(s) of the uploaded data.
```
guenthermi, line 205: I am not sure, if it is good to move the documentation of the arguments into this private function. Maybe it is better to move it into the public `create_run` function, where it is more visible and it is also the only point where this function is used.

guenthermi (commented):
Add to small comments

finetuner/experiment.py
jupyterjazz: not sure what you mean.. `create_run` still has the documentation of its arguments.

jemmyshin (approved):
LGTM!

finetuner/experiment.py
```diff
@@ -117,6 +120,79 @@ def delete_runs(self):
...
+        :param train_data: Either a `DocumentArray` for training data or a
+            name of the `DocumentArray` that is pushed on Hubble.
+        :param run_name: Name of the run.
+        :param train_data: Either a `DocumentArray` for evaluation data or a
+            name of the `DocumentArray` that is pushed on Hubble.
+        :param eval_callback: Evaluation callback that contains query and index data
+            and gets modified in-place.
+        :return: Name(s) of the uploaded data.
+        """
+        train_data_name = push_data_to_hubble(
```
gmastrapas, line 146: We make too many assumptions in this code block. Query data should be required. If not provided raise a value error. The only assumption we should make is that if index data are not given then query data == index data. And you don't even need to implement this. It is happening in the run

gmastrapas (changes requested):

finetuner/callbacks.py
```diff
@@ -117,3 +119,34 @@ class EarlyStopping:
...
+    :param limit: The number of top search results to consider when computing the
+        evaluation metrics.
+    :param distance: The type of distance metric to use when matching query and
+        index docs, available options are ``'cosine'``, ``'euclidean'`` and
+        ``'sqeuclidean'``.
+    """
+
+    query_data: Union[DocumentArray, str]
+    index_data: Optional[Union[DocumentArray, str]] = None
+    batch_size: Optional[int] = 8
```
gmastrapas, line 148: int

finetuner/experiment.py
```diff
@@ -117,6 +120,75 @@ def delete_runs(self):
...
+            else:
+                eval_callback.index_data = push_data_to_hubble(
+                    client=self._client,
+                    data=eval_callback.index_data,
+                    data_type=INDEX_DATA,
+                    experiment_name=self._name,
+                    run_name=run_name,
+                )
+        return train_data_name, eval_data_name
+
```
gmastrapas, line 191: The only thing that should be done here is to check whether eval, query and index data are given and if they are docarrays or strings. If they are docarrays they should be pushed.
```suggestion
        eval_data_name = None
        query_data_name = None
        index_data_name = None
        if eval_data:
            eval_data_name = push_data_to_hubble(
                client=self._client,
                data=eval_data,
                data_type=EVAL_DATA,
                experiment_name=self._name,
                run_name=run_name,
            )
         if query_data:
            query_data_name = push_data_to_hubble(
                client=self._client,
                data=query_data,
                data_type=QUERY_DATA,
                experiment_name=self._name,
                run_name=run_name,
            )
         if index_data:
            index_data_name = push_data_to_hubble(
                client=self._client,
                data=index_data,
                data_type=INDEX_DATA,
                experiment_name=self._name,
                run_name=run_name,
            )
        
        return train_data_name, eval_data_name, query_data_name, index_data_name

```

gmastrapas (changes requested):

finetuner/experiment.py
jupyterjazz: and we don't care that the same data could possibly be uploaded several times?

gmastrapas (approved):

PR #428, opened 2022-06-07
feat: add support for MLP model

gmastrapas:


finetuner/__init__.py
```diff
@@ -124,6 +126,7 @@ def fit(
     :param run_name: Name of the run.
     :param description: Run description.
     :param experiment_name: Name of the experiment.
+    :param model_options: Additional arguments to pass to the model construction.
```
bwanglzu, line 129: can we explain here what we could use? at least user can check developer reference to see what we can provide

finetuner/models.py
```diff
@@ -0,0 +1,158 @@
...
+    name = 'mlp'
+    description = 'Simple MLP encoder trained from scratch'
+    task = 'any'
+    output_dim = '-'
+    architecture = 'MLP'
+
+    def __init__(
+        self,
+        input_size: int,
+        hidden_sizes: Optional[List[int]] = None,
```
jemmyshin, line 56: I think `hidden_size` is not `Optional`. If `hidden_sizes` is None, what will be the output?

jemmyshin (changes requested):

finetuner/__init__.py
```diff
@@ -34,77 +36,82 @@ def login():
...
+        ]
+        for name, _model_class in _list_models().items()
+    }
+
+
+def describe_models() -> None:
+    """Describe available models in a table."""
     table = Table(title='Finetuner backbones')
+    header = models._ModelStub.header()
+    _no_wraps = [True] * (len(header) - 1) + [False]
```
bwanglzu, line 88: let's always `no_wrap=True`, i figured out it's not shwoing up correctly in jupyter notebook

finetuner/models.py
```diff
@@ -0,0 +1,158 @@
+import abc
+from typing import Any, Dict, List, Optional, Tuple, TypeVar
+
+ModelStubType = TypeVar('ModelStubType', bound='_ModelStub')
+
+
+class _ModelStub(metaclass=abc.ABCMeta):
```
bwanglzu, line 7: just curious, this base stub class has been excluded from the inspect, right?

finetuner/models.py
```diff
@@ -0,0 +1,158 @@
...
+    """ResNet50 model stub."""
+
+    name = 'resnet50'
+    description = 'Pretrained on ImageNet'
+    task = 'image-to-image'
+    output_dim = '2048'
+    architecture = 'CNN'
+
+    def __init__(self):
+        super(ResNet50, self).__init__()
```
bwanglzu, line 80: can we unify the coding style as MLP? `super().__init__(name=name, ...)`

bwanglzu (changes requested):
left some comments

finetuner/models.py
gmastrapas: yes https://github.com/jina-ai/finetuner/pull/428/files#diff-8de89b78deef000f7d454bc8ec2e228d003253f07319dbd3e455eeb18da24a77R53

finetuner/models.py
gmastrapas: what do you mean?

finetuner/models.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def __init__(self):
+        super(BERT, self).__init__()
+
+
+class SentenceTransformer(_ModelStub):
+    """SentenceTransformer model stub."""
+
+    name = 'sentence-transformers/msmarco-distilbert-base-v3'
+    description = 'Pretrained on Bert, fine-tuned on MS Marco'
```
violenil, line 152: Maybe 'Pretrained Bert model, fine-tuned on MS Marco' since Bert is not a dataset

finetuner/models.py
```diff
@@ -0,0 +1,158 @@
...
+    name = 'mlp'
+    description = 'Simple MLP encoder trained from scratch'
+    task = 'any'
+    output_dim = '-'
+    architecture = 'MLP'
+
+    def __init__(
+        self,
+        input_size: int,
+        hidden_sizes: List[int],
```
jemmyshin, line 56: Sorry, I rethink it a little bit, and maybe this will be better? 
```
hidden_sizes: Optional[List[int]] = None,
output_size: int,
```
Users can use only one layer / multi-layers of MLP.
`output_size` is demanded since it's the output. However, `hidden_sizes` will only be used in multi-layers cases, so it's `Optional`.

For short, `hidden_layer` will not be directly connected to input or output, so it's hidden by the model.

jemmyshin (changes requested):
LGTM!

jupyterjazz (approved):
LGTM!

finetuner/models.py
gmastrapas: We removed output_size since it can be mixed with output_dim that is a tailor arg

bwanglzu (approved):
LGTM!

PR #430, opened 2022-06-08
docs: add image-to-image example

jupyterjazz:
Example on fine-tuning `resnet50` for image-to-image retrieval task. 

[Link](https://ft-docs-resnet-guide--jina-docs.netlify.app/docs-resnet-guide/3_finetuner_in_action/3_2_image_to_image/)


docs/3_finetuner_in_action/3_2_image_to_image.md
```diff
@@ -1,3 +1,117 @@
 # Image to image search using ResNet
 
-Coming soon!
\ No newline at end of file
+This guide will show you how to fine-tune a ResNet model for image-to-image retrieval.
```
violenil, line 3: Perhaps we should standardize whether we want to hyphenate `image-to-image` or write instead `image to image` and make it consistent over all docs, including `text to text` etc.

docs/3_finetuner_in_action/3_2_image_to_image.md
```diff
@@ -1,3 +1,117 @@
...
+import finetuner
+finetuner.login()
+```
+
+## Choosing the model
+Now let's see what backbone models we can use. We can see available models either in [the docs](../2_step_by_step/2_5_choose_back_bone.md) or by calling:
+
+```python
+finetuner.describe_models()
+```
```
violenil, line 45: Would it make sense to show the output of this here in a new block?

docs/3_finetuner_in_action/3_2_image_to_image.md
```diff
@@ -1,3 +1,117 @@
...
+```{admonition} finetuner.fit parameters
+:class: tip
+The only mandatory parameters are `model` and `train_data`. We provide default values for others. Here is the [full list of the parameters](../../api/finetuner/#finetuner.fit). 
+```
+As you can see, we have to provide the `model` which we picked before. We also set `run_name` and `description`, which are optional,
+but recommended in order to retrieve your run easily and have some context about it. Furthermore, we had to provide `train_data` and `eval_data`. As you can see,
+we used the names of the `DocumentArray`s that are already on Hubble, but we could also pass a `DocumentArray` object itself, which will be automatically uploaded to Hubble. As stated before, we want to use the `TripletLoss`, and that's what `loss='TripletMarginLoss'` corresponds to.
+We used `BestModelCheckpoint` callback and set `epochs` to 6. The only difference between these two runs comes with the parameter `freeze`. As explained in the `description` field of each run, `freeze=True` will freeze the existing weights, and train only the additional `MLP` layer, while `freeze=False` will fine-tune the whole model.
```
violenil, line 87: Do you need to specify the size of the MLP layer that should be added? Or does `freeze=True` suffice?

violenil (commented):
I like this doc a lot! Just left some small comments/questions

docs/3_finetuner_in_action/3_2_image_to_image.md
jupyterjazz: Good point. Let's stick with `image to image`, `text to text`, etc.

docs/3_finetuner_in_action/3_2_image_to_image.md
jupyterjazz: I think you're right. `output_dim` needs to be specified. I'll check it though.

jemmyshin (approved):
LGTM!

docs/3_finetuner_in_action/3_2_image_to_image.md
```diff
@@ -1,3 +1,115 @@
...
+        train_data='resnet-ttl-train-data',
+        eval_data='resnet-ttl-eval-data',
+        loss='TripletMarginLoss',
+        callbacks=[BestModelCheckpoint()],
+        epochs=6,
+        freeze=True,
+        output_dim=2048,
+    )
+```
+Now, let's understand what this code means.
```
gmastrapas, line 95: this piece of code does

docs/3_finetuner_in_action/3_2_image_to_image.md
```diff
@@ -1,3 +1,115 @@
...
+        callbacks=[BestModelCheckpoint()],
+        epochs=6,
+        freeze=True,
+        output_dim=2048,
+    )
+```
+Now, let's understand what this code means. 
+```{admonition} finetuner.fit parameters
+:class: tip
+The only mandatory parameters are `model` and `train_data`. We provide default values for others. Here is the [full list of the parameters](../../api/finetuner/#finetuner.fit).
```
gmastrapas, line 98: required arguments

gmastrapas (changes requested):
We also need the evaluation results and how these two runs compare

docs/3_finetuner_in_action/3_2_image_to_image.md
```diff
@@ -1,3 +1,154 @@
...
+           DEBUG    Metric: 'model_r_precision' Value: 0.30105                                           __main__.py:217
+           DEBUG    Metric: 'model_recall_at_k' Value: 0.34500                                           __main__.py:217
+           DEBUG    Metric: 'model_reciprocal_rank' Value: 0.30105                                       __main__.py:217
+           INFO     Done ✨                                                                              __main__.py:219
+           INFO     Saving fine-tuned models ...                                                         __main__.py:222
+           INFO     Saving model 'model' in /usr/src/app/tuned-models/model ...                          __main__.py:233
+           INFO     Pushing saved model to Hubble ...                                                    __main__.py:240
+[10:38:14] INFO     Pushed model artifact ID: '62a1af491597c219f6a330fe'                                 __main__.py:246
+           INFO     Finished 🚀                                                                          __main__.py:248
+```
```
bwanglzu, line 154: in general when you submit 2 runs, I'm expecting `run2` should be a bit more performant than `run1`, while only reporting one metric seems to be a bit confused.

docs/3_finetuner_in_action/3_2_image_to_image.md
```diff
@@ -1,3 +1,154 @@
...
+│                                        resnet152 │ image-to-image │       2048 │          CNN │                               Pretrained on ImageNet │
+│                                  efficientnet_b0 │ image-to-image │       1280 │          CNN │                               Pretrained on ImageNet │
+│                                  efficientnet_b4 │ image-to-image │       1280 │          CNN │                               Pretrained on ImageNet │
+│                     openai/clip-vit-base-patch32 │  text-to-image │        768 │  transformer │ Pretrained on millions of text image pairs by OpenAI │
+│                                  bert-base-cased │   text-to-text │        768 │  transformer │       Pretrained on BookCorpus and English Wikipedia │
+│ sentence-transformers/msmarco-distilbert-base-v3 │   text-to-text │        768 │  transformer │           Pretrained on Bert, fine-tuned on MS Marco │
+└──────────────────────────────────────────────────┴────────────────┴────────────┴──────────────┴──────────────────────────────────────────────────────┘
+```
```
bwanglzu, line 59: not sure if it is needed, in every page, maybe we can just refer to the previous step-by-step secant. also explain a bit in this case, we're working on image to iamge search, so we can take a CNN model for fine-tuning

docs/3_finetuner_in_action/3_2_image_to_image.md
```diff
@@ -1,3 +1,154 @@
...
+        model='resnet50',
+        run_name='resnet-ttl-2',
+        description='freeze all weights, and only fine-tune the additional MLP layer.',
+        train_data='resnet-ttl-train-data',
+        eval_data='resnet-ttl-eval-data',
+        loss='TripletMarginLoss',
+        callbacks=[BestModelCheckpoint(), EvaluationCallback(query_data='resnet-ttl-eval-data')],
+        epochs=6,
+        learning_rate=0.01,
+    )
```
bwanglzu, line 92: when creating 2nd runs, please use a better set of configurations, learning rate `0.01` seems not quite reasonable to me

bwanglzu (changes requested):
left some comments

docs/3_finetuner_in_action/3_2_image_to_image.md
jupyterjazz: yeah, good point. Let's stick with one run then and make it two when we add a proper evaluation handling.

docs/3_finetuner_in_action/3_2_image_to_image.md
```diff
@@ -1,3 +1,124 @@
...
+You can either pull the data:
+```python
+from docarray import DocumentArray
+train_data = DocumentArray.pull('resnet-ttl-train-data')
+eval_data = DocumentArray.pull('resnet-ttl-eval-data')
+```
+Or specify given `DocumentArray` names (`resnet-ttl-train-data` and `resnet-ttl-eval-data`) directly to Finetuner.
+
+## Login to Finetuner
+As explained in the [Login to Jina ecosystem](../2_step_by_step/2_3_login_to_jina_ecosystem.md) section, first we need to login to Finetuner:
```
gmastrapas, line 34: login to Jina Cloud

gmastrapas (approved):
Looks good!

bwanglzu (approved):
LGTM!

PR #436, opened 2022-06-09
test: refactor flaky integration test

bwanglzu:


tests/integration/test_runs.py
```diff
@@ -38,84 +42,36 @@ def test_runs(
...
+        batch_size=10,
         epochs=1,
     )
     # list all runs
     runs = finetuner_mocker.list_runs(experiment_name=experiment_name)
     assert len(runs) == 2
-    run_names = sorted([run.name for run in runs])
-    assert run_names == [first_run, second_run]
-    # delete the first run
+    assert first_run and second_run in [run.name for run in runs]
```
jupyterjazz, line 54:
```suggestion
    assert first_run in [run.name for run in runs] and second_run in [run.name for run in runs]
```

jupyterjazz (commented):
I don't think this is a fix for the flaky test.
The problem is still the same. I think what you did was just polishing the integration tests. 

While one user is waiting for their run to finish, another user starts CI which calls `delete_experiments()` which deletes every existing experiment and run, thus when the first user is about to download the model, the run doesn't exist. 

I still like this refactoring though, I was thinking about it as well. But I don't think this changes the flaky test situation in any way.

CHANGELOG.md
```diff
@@ -48,6 +48,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Enable saving models for clip without overwriting. ([#432](https://github.com/jina-ai/finetuner/pull/432))
 
+- Fixed flaky integration test. ([#436](https://github.com/jina-ai/finetuner/pull/436))
```
gmastrapas, line 51: Fix

bwanglzu:
As discussed, we'll not test `delete_experiments` in IT. But only delete experiment by their name.

tests/integration/conftest.py
```diff
@@ -41,4 +41,4 @@ def get_auth_token():
 
     finetuner.login()
 
-    return finetuner.ft
+    yield finetuner.ft
```
jupyterjazz, line 44: Should we do `return` because we don't follow up with anything?

tests/integration/test_experiments.py
```diff
@@ -1,28 +1,32 @@
...
     assert exp1.status == 'ACTIVE'
     # create another experiment and list all experiments
     finetuner_mocker.create_experiment(second_exp_name)
-    exps = finetuner_mocker.list_experiments()
-    assert len(exps) == 2
-    experiment_names = sorted([exp.name for exp in exps])
-    assert experiment_names == [first_exp_name, second_exp_name]
-    assert exps[0].status == exps[1].status == 'ACTIVE'
+    experiments = finetuner_mocker.list_experiments()
+    assert len(experiments) == 3
```
jupyterjazz, line 18: If we check lengths, we still leave some space for failure. Another CI run might create its own experiments and overall it might be more than 3. Although I'm not sure how frequent this will be so we can leave it this way and observe.

tests/integration/test_runs.py
```diff
@@ -38,84 +42,37 @@ def test_runs(
...
-    runs = finetuner_mocker.list_runs(experiment_name=first_exp)
-    assert len(runs) == 1
-    assert runs[0].name == first_run
-    # delete experiments
-    finetuner_mocker.delete_experiments()
-    exps = finetuner_mocker.list_experiments()
-    assert not exps
+    finetuner_mocker.delete_experiment(experiment_name)
+    experiments = finetuner_mocker.list_experiments()
+    assert not experiments
```
jupyterjazz, line 67: same here in `assert not experiments`. you deleted your experiments but other run might have created some.

jupyterjazz (approved):
Added some comments. We still leave some chance for failure but I think it's still much more robust than what we had before. I don't really have a strong opinion about removing length checks now. We can leave it and see how frequently tests will fail because of that.

tests/integration/conftest.py
bwanglzu: makes no too much difference

tests/integration/test_runs.py
bwanglzu: now i only check if the created experiments in this list, good one.

tests/helper.py
```diff
@@ -0,0 +1,8 @@
+import random
+import string
+
+
+def create_random_name(prefix='experiment', length=6):
+    return f'{prefix}-' + ''.join(
+        random.choices(string.ascii_uppercase + string.digits, k=length)
+    )
```
jupyterjazz, line 8: why not use existing `get_random_name()`? and why do we need a prefix here?

PR #464, opened 2022-06-22
Chore bump hubble python client version

guenthermi:


guenthermi:
can not be merged because the API tests do not work with the new version.

PR #468, opened 2022-06-29
chore: remove time column from table

bwanglzu:
This PR removes `time` column in the readme table, which makes table shows correct and this column is useless since it has a linear relationship against training data size and device.


CatStark (approved):
Better! 👍🏼

PR #527, opened 2022-09-08
fix: create client automatically if user is logged in

guenthermi:
Create client automatically if user is already logged in
---

If the user is already logged in, finetuner can be used without calling `finetuner.login()`. This is especially helpful for people who want to integrate finetuner with other jina libraries that use hubble. Moreover, I add some useful exceptions if the user is not logged in and tries to call some functions which require a login. 


bwanglzu (approved):
LGTM!

PR #549, opened 2022-09-23
Fix log messages

guenthermi:
- Removes info and debug log messages from commons packages
- add spinner for waiting until finetuning job is started
- remove envelops in stream log messages
---

- [ ] I have added a line about this change to CHANGELOG

### DIFF
--- a/finetuner/__init__.py
+++ b/finetuner/__init__.py
@@ -1,4 +1,5 @@
 import inspect
+import logging
 import os
 import warnings
 from typing import Any, Dict, List, Optional, Union
@@ -346,6 +347,9 @@ def get_model(
       please install finetuner[full] to include all the dependencies.
     """
     from commons.data.inference import ONNXRuntimeInferenceEngine
+
+    for key in logging.root.manager.loggerDict:
+        logging.getLogger(key).setLevel('CRITICAL')
 
     if gpu:
         warnings.warn(
--- a/finetuner/client/client.py
+++ b/finetuner/client/client.py
@@ -186,7 +186,10 @@ class FinetunerV1Client(_BaseClient):
         response = self._handle_request(url=url, method=GET, stream=True)
         for entry in response.iter_lines():
             if entry:
-                yield entry.decode('utf-8', errors='ignore')
+                decoded_message: str = entry.decode('utf-8', errors='ignore')
+                if decoded_message.startswith('data: '):
+                    decoded_message = decoded_message[6:]
+                yield decoded_message
 
     def create_run(
         self,
--- a/finetuner/client/client.py
+++ b/finetuner/client/client.py
@@ -187,9 +187,14 @@ class FinetunerV1Client(_BaseClient):
         for entry in response.iter_lines():
             if entry:
                 decoded_message: str = entry.decode('utf-8', errors='ignore')
-                if decoded_message.startswith('data: '):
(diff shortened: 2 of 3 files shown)


finetuner/run.py
```diff
@@ -81,17 +83,18 @@ def stream_logs(self, interval: int = 5) -> Iterator[str]:
...
-            status = self.status()[STATUS]
-            if status == CREATED:
-                msg = (
-                    f'Preparing to run, logs will be ready to pull when '
-                    f'`status` is `STARTED`. Current status is `{status}`'
-                )
-                print(msg)
-            else:
-                break
+        console = Console(width=200)
```
bwanglzu, line 86: create `Console` outside from the function, also check `describe_models`.

finetuner/client/client.py
```diff
@@ -186,7 +186,15 @@ def stream_run_logs(self, experiment_name: str, run_name: str) -> Iterator[str]:
         response = self._handle_request(url=url, method=GET, stream=True)
         for entry in response.iter_lines():
             if entry:
-                yield entry.decode('utf-8', errors='ignore')
+                decoded_message: str = entry.decode('utf-8', errors='ignore')
+                sep_pos = decoded_message.find(': ')
+                if sep_pos != -1:
```
bwanglzu, line 191: can you elaberate what is this doing?

bwanglzu (changes requested):
i like the spiner! much better than print

finetuner/client/client.py
guenthermi: The EventSourceResponse class generates different kinds of log message (e.g., "event" and "data" messages). It adds this type at the beginning of the message, e.g., "data: Finetuning ...". This code snippet divides the type and the message by splitting it at the first occurrence of ": " and only returns the message itself as log message.

bwanglzu: what if `sep_pos` is None?

bwanglzu (commented):
minor comment, can you attach a screenshot in the PR description of the spinner?

finetuner/client/client.py
guenthermi: The `find` function returns -1 if it is not present. Therefore, I check for -1

finetuner/__init__.py
```diff
@@ -347,6 +326,9 @@ def get_model(
     """
     from commons.data.inference import ONNXRuntimeInferenceEngine
 
+    for key in logging.root.manager.loggerDict:
+        logging.getLogger(key).setLevel('CRITICAL')
```
gmastrapas, line 330: why are you silencing all loggers?

guenthermi: I thought that we only want to log the messages of the executor which are set by the ONNXRuntimeInferenceEngine. However, I can also change it to only silence the commons and the ulrlib logger in the case we want to log more. However, this means that there might occur other unexpected log messages after some changes.

gmastrapas: try to silence only these loggers first, and see how it looks

guenthermi:
Cleaned up loggs:
Screenshot from 2022-09-27 14-32-20
Spinner which shows up before logs are available:
Screenshot from 2022-09-27 14-29-07

bwanglzu (approved):
LGTM!

PR #560, opened 2022-09-28
feat: add torch inference

bwanglzu:
This PR allows us to use torch model for inference, most of the job has been done in core.

- [x] adopt configuration changes in core
- [x] bump commons package once core released
- [x] documentation update

```python
run = finetuner.fit(
    ...,
    ...,
    ...,
    to_onnx=True,
)

# inference with finetuner client default `pt`
model = finetuner.get_model(..., is_onnx=True)
finetuner.encode(model, da)
# inference with finetuner executor, default `pt`
FinetunerExecutor(..., uses_with={'is_onnx': True})
```
---


setup.py
```diff
@@ -68,7 +68,7 @@
         ],
         extras_require={
             'full': [
-                'finetuner-commons==0.10.1',
+                'finetuner-commons==0.0.1b8',
```
gmastrapas, line 71: did you push this to pypi? then tests should be working right?

gmastrapas (approved):

setup.py
bwanglzu: some of the metadata is missing when validating, such as `builder`, so it will fail until release finished

guenthermi (approved):
LGTM

PR #577, opened 2022-10-16
refactor: unify cpu gpu into device

bwanglzu:
Resolve inconsistency between `fit` end `get_model`, in `fit` we use `cpu`, in `get_model` we use `gpu`. Better to align with Docarray `device='cpu'` or `device='cuda'`.

This will not be a breaking change for internal users. 

This will be a breaking change for ppl install `finetuner-full`  when calling `get_model`.

A follow up ticket has been created to remove `cpu` in `0.7.0` #578 

```diff
# training time
run = finetuner.fit(
    ...,
-   cpu=False,  # cpu is still supported, no breaking change, we sunset `cpu` after 0.7
+   device='cuda', # note, now `cuda` is the default device.
)

# inference time
inference_engine = finetuner.get_model(
     ...,
-    gpu=True, 
+    device='cuda',  # note, `cpu` is the default device unless user specify `cuda`.
)
```

---

### DIFF
--- a/finetuner/__init__.py
+++ b/finetuner/__init__.py
@@ -110,8 +110,8 @@ def fit(
     scheduler_step: str = 'batch',
     freeze: bool = False,
     output_dim: Optional[int] = None,
-    cpu: bool = True,
-    device: str = 'cpu',
+    cpu: bool = False,
+    device: str = 'cuda',
     num_workers: int = 4,
     to_onnx: bool = False,
 ) -> Run:
--- a/finetuner/__init__.py
+++ b/finetuner/__init__.py
@@ -168,6 +168,7 @@ def fit(
     :param output_dim: The expected output dimension as `int`.
         If set, will attach a projection head.
     :param cpu: Whether to use the CPU. If set to `False` a GPU will be used.
+        Will be deprecated from 0.7.0.
     :param device: Whether to use the CPU, if set to `cuda`, a Nvidia GPU will be used.
         otherwise use `cpu` to run a cpu job.
     :param num_workers: Number of CPU workers. If `cpu: False` this is the number of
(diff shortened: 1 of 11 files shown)


finetuner/experiment.py
```diff
@@ -165,14 +167,24 @@ def create_run(
             **kwargs,
         )
 
-        cpu = kwargs.get(CPU, True)
+        device = kwargs.get(DEVICE, 'cuda')
+        if device == 'cpu' and not kwargs.get(CPU, True):
```
gmastrapas, line 171: can you explain the logic here?

gmastrapas: so it should be `if device == 'cuda' and kwargs.get(CPU, True):`?

bwanglzu: yes it's more clear

gmastrapas (changes requested):

finetuner/__init__.py
```diff
@@ -338,7 +347,7 @@ def get_model(
         TorchInferenceEngine,
     )
 
-    if gpu:
+    if device == 'cuda' and is_onnx:
         warnings.warn(
             message='You are using cuda device for ONNX inference, please consider'
             'call `pip install onnxruntime-gpu` to speed up inference.',
```
gmastrapas, line 353: consider calling

finetuner/__init__.py
```diff
@@ -167,10 +168,16 @@ def fit(
     :param output_dim: The expected output dimension as `int`.
         If set, will attach a projection head.
     :param cpu: Whether to use the CPU. If set to `False` a GPU will be used.
```
gmastrapas, line 170: can you add here that this arg will be deprecated, and you should use `device` instead?

gmastrapas (changes requested):

LMMilliken (approved):

gmastrapas (approved):

PR #598, opened 2022-11-04
feat: add support for raw data when encoding

LMMilliken:
This pr adds support for providing a list of strings instead of a DocumentArray when using the `finetuner.encode` function. The modality of the contents of this list can be inferred using the model provided to the function, so no additional arguments need to be provided.
The docs have been updated to reflect this change, however encoding with a list has not been given priority over encoding with a DocumentArray. Example code snippets are left unchanged, and some additional code snippets are added.

---


finetuner/__init__.py
```diff
@@ -485,3 +488,5 @@ def encode(
             inputs = model._move_to_device(inputs)
             output = model.run(inputs)
             batch.embeddings = output.detach().cpu().numpy()
+
+    return data
```
LMMilliken, line 492: Since the encode method previously worked on the input `DocumentArray` in place, I was not sure how to return the embeddings in the case that the user provided a list. Currently I am just returning the final DocumentArray in both cases, though I would welcome alternative suggestions

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add support for CSV files to the `fit` function. [#592](https://github.com/jina-ai/finetuner/pull/592)
 
+- Add support for lists to the `encode` function. [#598](https://github.com/jina-ai/finetuner/pull/598)
+
+
```
gmastrapas, line 17: one new line between sections of the same release

finetuner/__init__.py
```diff
@@ -448,7 +449,7 @@ def get_model(
 
 
 def encode(
-    model,
+    model: 'InferenceEngine',
     data: DocumentArray,
     batch_size: int = 32,
 ):
```
gmastrapas, line 455: specify a return type as well

finetuner/data.py
```diff
@@ -58,6 +73,34 @@ def build_dataset(
     return data
 
 
+def build_encoding_dataset(
+    model: 'InferenceEngine', data: Union[List[str], DocumentArray]
```
gmastrapas, line 77: just to be conistent with the other method you can accept a str here as model

finetuner/data.py
```diff
@@ -58,6 +73,34 @@ def build_dataset(
...
+) -> DocumentArray:
+    """If data has been provided as a list, a :class:`DocumentArray` is created
+        from the elements of the list
+
+    :param model: The InferenceEngine for the model being used for encoding
```
gmastrapas, line 82: dont document these functions with arguments. They are internal. Just a docstring is fine

finetuner/data.py
```diff
@@ -39,12 +42,24 @@ class CSVOptions:
...
-def build_dataset(
+def build_finetuning_dataset(
     data: Union[str, TextIO, DocumentArray],
     model: str,
     csv_options: Optional[CSVOptions] = None,
 ) -> Union[str, DocumentArray]:
+    """If data has been provided as a CSV file, the given CSV file is parsed
+        and a :class:`DocumentArray` is created.
+
+    :param data: The input data provided by the user, either a path to a CSV file,
```
gmastrapas, line 53: dont document these functions with arguments. They are internal. Just a docstring is fine

gmastrapas (changes requested):

README.md
```diff
@@ -187,6 +187,20 @@ finetuner.encode(model=model, data=da)
 da.summary()
 ```
 
+When encoding, you can provide data either as a DocumentArray or a list. Since the modality of your input data can be inferred from the model being used, there is not need to provide any additional information besides the content you want to encode:
```
guenthermi, line 190:
```suggestion
When encoding, you can provide data either as a DocumentArray or a list. Since the modality of your input data can be inferred from the model being used, there is no need to provide any additional information besides the content you want to encode:
```

Maybe you can put the second sentence in an info / hint box below the code.

docs/walkthrough/integrate-with-jina.md
```diff
@@ -59,6 +59,25 @@ Text of the returned document: some text to encode
 Shape of the embedding: (768,)
 ```
 
+### Encoding a List
+Data that is stored in a regular list can be embedded in the same way you would a [DocumentArray](https://docarray.jina.ai/). Since the modality of your input data can be inferred from the model being used, there is not need to provide any additional information besides the content you want to encode:
```
guenthermi, line 63:
```suggestion
Data that is stored in a regular list can be embedded in the same way you would a [DocumentArray](https://docarray.jina.ai/). Since the modality of your input data can be inferred from the model being used, there is no need to provide any additional information besides the content you want to encode:
```
same here

finetuner/__init__.py
```diff
@@ -448,10 +449,10 @@ def get_model(
 
 
 def encode(
-    model,
+    model: 'InferenceEngine',
     data: DocumentArray,
```
guenthermi, line 453: if it is a list it is not a DocumentArray anymore

finetuner/data.py
```diff
@@ -58,6 +63,28 @@ def build_dataset(
     return data
 
 
+def build_encoding_dataset(
+    model: Union['InferenceEngine', str], data: Union[List[str], DocumentArray]
+) -> DocumentArray:
+    """If data has been provided as a list, a :class:`DocumentArray` is created
+    from the elements of the list
+    """
```
guenthermi, line 71: The parameter descriptions are missing

finetuner/data.py
```diff
@@ -58,6 +63,28 @@ def build_dataset(
...
+    if not isinstance(data, DocumentArray):
+        try:
+            task = get_stub(model._model_name, select_model=model._select_model).task
+            if task == 'any':
+                raise ValueError('MLP model does not support values from a list.')
+            task = task.split('-to-')[0]
+        except Exception:
+            task = model._model_name.split('-')[1]
```
guenthermi, line 79: I think here is something wrong. When a clip model is used, you can encode text as well as images with it. So it can either be text or image and you can not infer this from the task. Moreover, I don't understand what the except block does. If you pass any string to `test`, `test.split('-to-')[0]` will always just return the text.

finetuner/data.py
LMMilliken: @gmastrapas suggested they be removed as this is an internal function.

README.md
```diff
@@ -187,6 +187,20 @@ finetuner.encode(model=model, data=da)
...
+
+```python
+import finetuner
+from docarray import Document, DocumentArray
+
+da = ['~/Pictures/your_img.png']
```
bwanglzu, line 196: let's avoid `da` as name since it's short for document array, something like `images` or `imgs` or `img_urls` would be better

docs/walkthrough/integrate-with-jina.md
```diff
@@ -59,6 +59,25 @@ Text of the returned document: some text to encode
...
+
+model = finetuner.get_model('/path/to/YOUR-MODEL.zip')
+
+lst = ['some text to encode']
+
+da = finetuner.encode(model=model, data=lst)
+
+for doc in da:
+    print(f'Text of the returned document: {doc.text}')
+    print(f'Shape of the embedding: {doc.embedding.shape}')
```
bwanglzu, line 77: if user provide a list of object, we should not convert it as a da, but an numpy ndarray of embeddings.

bwanglzu (changes requested):
this is a good point, if user want to encode a `DocumentArray`, we should return a `DocumentArray`, otherwise if user want to encode a list of objects, i believe its better to return a `np.ndarray` of embeddings.

docs/walkthrough/integrate-with-jina.md
LMMilliken: That makes sense, changed now

docs/notebooks/image_to_image.ipynb
```diff
@@ -1,23 +1,10 @@
 {
```
gmastrapas, line 1: what happend here? the notebook changed but not the md file?

finetuner/data.py
```diff
@@ -58,6 +63,35 @@ def build_dataset(
...
+    """
+    if not isinstance(data, DocumentArray):
+
+        modalities = model._metadata['preprocess_types']
+        if model._select_model:
+            task = modalities[model._select_model]
+        else:
+            task = list(modalities)[0]
+
+        # try:
```
gmastrapas, line 80: commented section should be removed

finetuner/data.py
```diff
@@ -58,6 +63,35 @@ def build_dataset(
...
+        #     task = get_stub(model._model_name, select_model=model._select_model).task
+        #     if task == 'any':
+        #         raise ValueError('MLP model does not support values from a list.')
```
gmastrapas, line 83: should we still raise an exception here?

docs/notebooks/image_to_image.ipynb
LMMilliken: ah, I didn't mean to commit that, i think since the notebook file records metadata such as number of runs, running the notebook causes changes that I committed mistakenly

LMMilliken: I attempted to revert the changes to that file to main but it has since been updated there as well, after merging there will be no difference to either the notebook or the markdown

finetuner/data.py
```diff
@@ -58,6 +63,29 @@ def build_dataset(
     return data
 
 
+def build_encoding_dataset(
+    model: Union['InferenceEngine', str], data: Union[List[str], DocumentArray]
```
gmastrapas, line 67: can this argument be a string?

LMMilliken: No it cannot, changed now

docs/walkthrough/integrate-with-jina.md
```diff
@@ -59,6 +59,25 @@ Text of the returned document: some text to encode
...
+### Encoding a List
+Data that is stored in a regular list can be embedded in the same way you would a [DocumentArray](https://docarray.jina.ai/). Since the modality of your input data can be inferred from the model being used, there is no need to provide any additional information besides the content you want to encode. When providing data as a list, the `finetuner.encode` method will return a `np.ndarray` of embeddings, instead of a `docarray.DocumentArray`:
+
+```python
+from docarray import DocumentArray, Document
+import finetuner
+
+model = finetuner.get_model('/path/to/YOUR-MODEL.zip')
+
+lst = ['some text to encode']
```
bwanglzu, line 71: i would rename it to `texts`, then below:
```python
for text, embedding in zip(texts, embeddings):
    ...
```

bwanglzu (approved):
LGTM!, Before release, let's also update notebooks `inference` session, and make inference from `list` as default option. see https://finetuner.jina.ai/notebooks/text_to_text/#inference

finetuner/__init__.py
```diff
@@ -469,6 +471,9 @@ def encode(
 
     from _finetuner.models.inference import ONNXRuntimeInferenceEngine
 
+    return_da = isinstance(data, DocumentArray)
+    data = build_encoding_dataset(model=model, data=data)
```
guenthermi, line 475: I would prefer to call this function only if the input type is a list.

guenthermi (approved):
LGTM, only added one minor comment

PR #623, opened 2022-11-28
Fix build encoding dataset

LMMilliken:
Currently, attempting to encode data from a list using a CLIP model made using `get_model` will result in an error, this pr prevents that by changing the method of determining the inferring the modality of the model

---


gmastrapas (approved):

bwanglzu (approved):
LGTM!

PR #628, opened 2022-12-01
ci: add remote-ci for core

LMMilliken:
This pr adds a CI job that can be manually triggered and will run tests on finetuner-core

currently, the core-ci workflow can be triggered in two ways:
A workflow dispatch, in which a user can manually trigger the workflow from the actions tab of this repo
A repository dispatch, in which the another repository (expected to be core) can remotely trigger the workflow.
Both of these methods only work when the workflow exists on the master branch, meaning that this pr needs to be tested before it can be merged

---

### DIFF
--- a/.github/workflows/remote-ci.yml
+++ b/.github/workflows/remote-ci.yml
@@ -30,7 +30,7 @@ jobs:
         with:
           repository: jina-ai/finetuner-core
           ref: main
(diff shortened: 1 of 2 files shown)


bwanglzu (approved):
LGTM! We still need to sent the event in core to trigger the event?

guenthermi (approved):
LGTM

PR #640, opened 2022-12-20
ci: authenticate repo-dispatch

LMMilliken:
Currently, the core-ci workflow will always run the core tests, regardless of who triggered the workflow. This pr requires that a github token be passed in the body of the repository dispatch that triggers the core-ci, this token is then used when pulling core to test. If the token is not valid, then core wont be pulled and the tests wont run

---


guenthermi (approved):

bwanglzu (approved):
LGTM!

PR #643, opened 2022-12-29
docs: refine documentation 0.7

bwanglzu:
---

- [ ] This PR references an open issue
- [ ] I have added a line about this change to CHANGELOG

PR #649, opened 2023-01-09
docs: add namespace to artifact names

guenthermi:
Add namespace to artifact names in the documentation
---
Updates all artifact names in the documentation and the README to include the namespace prefix `finetuner`.
Contains some changes from https://github.com/jina-ai/finetuner/pull/643, because the notbooks already contain them.

### DIFF
--- a/docs/notebooks/image_to_image.md
+++ b/docs/notebooks/image_to_image.md
@@ -7,8 +7,7 @@ jupyter:
       format_version: '1.3'
       jupytext_version: 1.14.1
   kernelspec:
-    display_name: 'Python 3.7.15 (''.venv'': venv)'
-    language: python
+    display_name: Python 3
     name: python3
 ---
 
@@ -61,9 +60,9 @@ finetuner.login(force=True)
 ```
 
 ```python id="ONpXDwFBsqQS"
-train_data = DocumentArray.pull('tll-train-data', show_progress=True)
-query_data = DocumentArray.pull('tll-test-query-data', show_progress=True)
-index_data = DocumentArray.pull('tll-test-index-data', show_progress=True)
+train_data = DocumentArray.pull('finetuner/tll-train-data', show_progress=True)
+query_data = DocumentArray.pull('finetuner/tll-test-query-data', show_progress=True)
+index_data = DocumentArray.pull('finetuner/tll-test-index-data', show_progress=True)
 
 train_data.summary()
 ```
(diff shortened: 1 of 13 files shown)


docs/notebooks/multilingual_text_to_image.md
```diff
@@ -12,62 +12,72 @@ jupyter:
...
 
-
+<!-- #region id="ed1e7d55-a458-4dfd-8f4c-eeb02521c221" -->
 ## Install
+<!-- #endregion -->
 
-```python
-!pip install 'finetuner[full]'
+```python id="9261d0a7-ad6d-461f-bdf7-54e9804cc45d"
+!pip install finetuner
```
LMMilliken, line 36: A code block towards the end uses the `encode` method, do we not need 'finetuner[full]' for that?

guenthermi: Good catch! Not sure why this was changes

docs/notebooks/image_to_image.md
```diff
@@ -244,13 +245,12 @@ for i, (doc_pt, doc_ft) in enumerate(zip(query_pt, query_ft)):
...
-
-On the other hand, there are also cases where the fine-tuned model performs worse, and fails to correctly match images that it previously could. This case is much rarer than the the previous case. For this dataset there were 108 occasions where the fine-tuned model returned the correct pair where it couldn't before, and oly 33 occasions where the the finetuned model returned an incorrect image after fine-tuning but returned a correct one before. Nevertheless it still can happen:
-
-![image-image-triplets-bad](images/image-image-triplets-bad.png)
-
+![image-image-triplets-good](https://finetuner.jina.ai/_images/image-image-triplets-good.png)
 
+On the other hand, there are also cases where the fine-tuned model performs worse, and fails to correctly match images that it previously could. This case is much rarer than the the previous case. For this dataset there were 108 occasions where the fine-tuned model returned the correct pair where it couldn't before, and only 33 occasions where the the finetuned model returned an incorrect image after fine-tuning but returned a correct one before. Nevertheless it still can happen:
```
gmastrapas, line 253:
```suggestion
On the other hand, there are also cases where the fine-tuned model performs worse, and fails to correctly match images that it previously could. This case is much rarer than the previous case. For this dataset there were 108 occasions where the fine-tuned model returned the correct pair where it couldn't before, and only 33 occasions where the finetuned model returned an incorrect image after fine-tuning but returned a correct one before. Nevertheless it still can happen:
```

gmastrapas (approved):

PR #656, opened 2023-01-17
docs: update artifact names in image to image notebook

guenthermi:
Update artifact names in image to image notebook
---

The artifact names of the image-to-image tutorial only worked without `finetuner/` prefix since they are not in the finetuner account. Howver, there the same artifacts are in the finetuner account with slightly different names. Therefore the names needed to be changed from `*-data` to `*-da`


guenthermi:
We instead pushed the correct artifacts with the names in the documentation

PR #663, opened 2023-01-23
chore: add example CSV files

LMMilliken:
This pr adds a folder containing several CSV files that are suitable to be uploaded to the JAC as data for a finetuning run.
It is worth noting the the file `image-text_or_image_example.csv` is suitable for both text-image and image-image tasks as each row can be interpreted as either a `image-text` pair (suitable for text-image) or a `image-label` pair (suitable for image-image)

---


LMMilliken:
The mesh-mesh example wont work on JAC at the moment, as they are not using the latest version of core, which is needed to support pointnet++

LMMilliken:
> Can you rename folder to `data`?

the name 'data' is in one of our gitignore files, i have renamed to example_data instead

guenthermi (approved):
LGTM

gmastrapas:
> > Can you rename folder to `data`?
> 
> the name 'data' is in one of our gitignore files, i have renamed to example_data instead

`datasets` then? I really dislike `_` names, especially in the root dir 🥲

PR #667, opened 2023-01-26
fix: add working example CSV for image-image

LMMilliken:
This pr adds an example CSV file for image-image models, previously one file was used for both image-image and text-image but the file cannot be parsed correctly without a `csv_options` parameter, which is not provided by JAC at the moment

---


guenthermi (approved):
LGTM, maybe add the current date as "access date" to the credits.

PR #664, opened 2023-01-24
feat: integrate new loss and pooling options

LMMilliken:
This pr adds support for the newly added loss and pooling options from core, as well as a new page in the documentation explaining how to use them.

---

- [x] This PR references an open issue
- [x] I have added a line about this change to CHANGELOG

PR #668, opened 2023-01-26
Feat support continue training

guenthermi:
Add support for continue the training
---

This PR enables to continue the training form an existing artifact of a fine-tuned model

```python
train_data = 'path/to/another/data.csv'

run2 = finetuner.fit(
    model='efficientnet_b0',
    train_data=train_data,
    model_artifact=run.artifact_id,
)
print(f'Run name: {run2.name}')
print(f'Run status: {run2.status()}')
```

Since we need the model name to detect the task in order to construct the training dataset correctly, users still need to add the model property to the `fit` function.


guenthermi:
Tests are failing because I did not commit any requirements change for commons and stubs yet.

guenthermi:
I added a pre-release to see if the tests are running. Everything seems to work as expected. One test is failing because loss-optimizers are not added yet. This should work when https://github.com/jina-ai/finetuner/pull/664 is merged.

docs/walkthrough/run-job.md
```diff
@@ -38,6 +38,29 @@ the run status changes from:
...
+## Continue training on new data
+
+Finetuner also supports continuing training a model produced by a previous fine-tuning run.
+If you have additional data which you want to use to further train a model, you can do this by passing its artifact id to the fit function:
+
+```python
+train_data = 'path/to/another/data.csv'
+
+run2 = finetuner.fit(
+    model='efficientnet_b0',
```
bwanglzu, line 50: do we still need the `model`?

bwanglzu: okay now i see the hint, do i understand correly that the `model` type must be identical to `artifact` model type?

bwanglzu (commented):
LGTM! while why CI is failing?

docs/walkthrough/run-job.md
```diff
@@ -38,6 +38,29 @@ the run status changes from:
...
+    model_artifact=run.artifact_id,
+)
+print(f'Run name: {run2.name}')
+print(f'Run status: {run2.status()}')
+```
+
+```{admonition} Continue training requires a model parameter 
+:class: hint
+When you want to continue training, you still need to provide the `model` parameter
+beside the `model_artifact` parameter for Finetuner to correctly configure the new run.
```
LMMilliken, line 61:
```suggestion
:class: hint
When you want to continue training, you still need to provide the `model` parameter
as well as the `model_artifact` parameter for Finetuner to correctly configure the new run.
```

LMMilliken, line 61:
```suggestion
:class: hint
When you want to continue training, you still need to provide the `model` parameter
as well as the `model_artifact` parameter for Finetuner to correctly configure the new run.
```

finetuner/experiment.py
```diff
@@ -237,7 +238,8 @@ def _create_config_for_run(
         ]
         public = kwargs[PUBLIC] if kwargs.get(PUBLIC) else False
         model = config.ModelConfig(
-            name=model,
+            name=model if not kwargs.get(MODEL_ARTIFACT) else None,
+            artifact=kwargs.get(MODEL_ARTIFACT),
```
LMMilliken, line 242: Just to make sure I understand, this means that if there is a `MODEL_ARTIFACT`, then `name` is not set?

LMMilliken (approved):
Once the pr for the new pooling and loss options is merged, everything should pass.
LGTM!

docs/walkthrough/run-job.md
guenthermi: We need the model name to detect the task, so the task they are used for must be equal, so basically yes. If you use a docarray dataset, it probably does not matter

guenthermi:
> LGTM! while why CI is failing?

as said in the comment above, the stubs version includes already the ArcFace changes and specifically, the loss-optimizers etc. The tests therefore failing but it will be addressed by Louis PR. So I need to merge main into this branch here after Louis finished its PR then it should not fail anymore

finetuner/experiment.py
guenthermi: yes, because George implemented the jsonschema in a way that both arguments are mutually exclusive, which makes somehow sense, because otherwise it would be a little bit hard to see what will happen, e.g., if a new model is constructed or the artifact is used. So in this way, the final config to be send to the API is more clear on what a run should do.

gmastrapas (approved):

PR #711, opened 2023-04-12
refactor: adjust schema based on job type

bwanglzu:
Since now we support data generation job, some adjustments in the client is needed.

---


LMMilliken (approved):

finetuner/experiment.py
```diff
@@ -287,7 +287,7 @@ def _create_config_for_run(
         if kwargs.get(LOSS_OPTIMIZER_OPTIONS):
             hyper_parameters.loss_optimizer_options = kwargs.get(LOSS_OPTIMIZER_OPTIONS)
 
-        run_config = config.RunConfig(
+        run_config = config.FinetuningConfig(
```
guenthermi, line 290: Maybe we should rename the function, this line is part of (_create_config_for_run) in something like _create_config_for_finetuning_run.

PR #738, opened 2023-04-23
docs: add callbacks

bwanglzu:
add back callbacks to new branch. With some minor updates:

1. rewrite the first paragraph.
2. update code examples and outputs.
---


docs/concepts/callbacks.md
```diff
@@ -1,2 +1,249 @@
...
+
+run = finetuner.fit(
+    ...,
+    callbacks=[
+        EvaluationCallback(
+            query_data='finetuner/tll-test-query-da',
+            index_data='finetuner/tll-test-index-da'
+        ),
+        WandBLogger(),
+    ]
```
LMMilliken, line 229: The callbacks are in the right order in this example but should we also explain that the evaluation callback has to come before the wandblogger in order for the evaluation metrics to be tracked?

guenthermi (commented):
Since we have no API reference for the callbacks it would be nice to at least list there options here. Maybe of the options are not even documented here.

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
 (callbacks)=
-# {octicon}`link` Callbacks
\ No newline at end of file
+# {octicon}`link` Callbacks
+
+Callbacks are a way of adding additional methods to the finetuning process.
+The methods are executed when certain events occur and there are several callback classes, each serving a different function by providing different methods for different events.
```
scott-martens, line 5:
```suggestion
A callback is a function that can optionally run when a specific kind of event occurs during a process. There are several events during a Finetuner run that can have callbacks. 
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+A run can be assigned multiple callbacks using the optional `callbacks` parameter when it is created.
```
scott-martens, line 6:
```suggestion
Users assign callbacks to the `finetuner.fit` method with the optional `callbacks`.
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+
+## EvaluationCallback
+
+The `EvaluationCallback` is used to calculate retrieval metrics for the model being tuned at the end of each epoch.
```
scott-martens, line 10:
```suggestion
The `EvaluationCallback` calculates retrieval metrics at the end of each epoch for the model being tuned.
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+In order to evaluate the model, two additional data sets - a query dataset and an index dataset - need to be provided as arguments.
+If no index set is provided, the query dataset is reused instead.
```
scott-martens, line 12:
```suggestion
If no index set is provided, the evaluation is performed with the query dataset.
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+
+It is worth noting that the evaluation callback and the `eval_data` parameter of the fit method do not do the same thing.
+The `eval_data` parameter is used to evaluate the loss of the model.
+On the other hand, the evaluation callback is used to evaluate the quality of the searches using metrics such as average precision and recall.
+These search metrics can be used by other callbacks if the evaluation callback is first in the list of callbacks when creating a run.
+
+```{admonition} Evaluation callback with two models
+:class: hint
+Usually, you don't need to provide the name of a model to the evalution callback.
+The callback just takes the model which is fine-tuned.
```
scott-martens, line 61:
```suggestion
You don't usually need to provide the name of a model to the evaluation callback. It assumes you mean to evaluate the model that you are fine-tuning.
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+However, if multiple models are involved in the fine-tuning process, like this is the case for CLIP models, it needs to be clear which model is used to encode the documents in `query_data` and `index_data`.
+This can be specified by the `model` attribute of the callback.
```
scott-martens, line 63:
```suggestion
However, if multiple models are involved in the fine-tuning process, for example, if you fine-tune CLIP models, then you need to be clear about which model to use to encode the documents in `query_data` and `index_data`.
This is specified by the `model` parameter of the callback.
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+If a different model should be used for the `index_data`, you can set this via the `index_model` attribute.
```
scott-martens, line 64:
```suggestion
If the `index_data` should be encoded by a different model from the `query_data`, you must specify this with the `index_model` parameter.
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+Alternatively, you can use the {func}`~finetuner.run.Run.display_examples()` function to display a table of the Top-K results before and after fine-tuning to the console.
+
+![Example Results](https://user-images.githubusercontent.com/6599259/224284912-f3f6f547-8d75-4529-8df1-7f9904423239.png)
+
+
+## BestModelCheckpoint
+
+This callback evaluates the performance of the model at the end of each epoch, and keeps a record of the best performing model across all epochs.
+Once fitting is finished the best performing model is saved instead of the most recent model.
+The definition of best is based on two parameters:
```
scott-martens, line 92:
```suggestion
Finetuner determines which model is the best is based on two parameters:
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+
+- `monitor`: The metric that is used to compare models to each other. By default this value is `val_loss`, the loss function calculated using the evaluation data, however the loss calculated on the training data can be used instead with `train_loss`; any metric that is recorded by the evaluation callback can also be used.
```
scott-martens, line 94:
```suggestion
- `monitor`: This parameter is by default `val_loss`, which uses the evaluation data to compare models. Alternatively, you can set this to `train_loss`, which will compare models using the training data. You can specify any metric recorded by the evaluation callback for this parameter.
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+- `mode`: Whether the monitored metric should be maximised (`max`) or minimised (`min`). By default the mode is set to `auto`, meaning that it will automatically choose the correct mode depending on the chosen metric: 'min' if the metric is loss and 'max' if the metric is one recorded by the evaluation callback.
+
+The console output below shows how the evaluation loss of the model is monitored between each epoch, and how the best performing model is tracked. Since the final model has a higher loss than the previously recorded best model, the best model will be saved instead of the latest one.
```
scott-martens, line 97:
```suggestion
The console output below shows how the evaluation loss of the model is monitored between epochs and how the best-performing model is tracked. Since the final model has a higher loss than the previously recorded best model, the best model will be saved instead of the latest one.
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+3. When alpha is a floating-point number between 0 and 1, we combine the weights of the pre-trained and fine-tuned models.
+
+```{warning}
+It is recommended to use WiSEFTCallback when fine-tuning CLIP.
+We can not ensure it works for other types of models, such as ResNet or BERT.
+
+Please refer to {ref}`Apply WiSE-FT <wise-ft>` in the CLIP fine-tuning example for more information.
+```
+
+## WandBLogger
```
scott-martens, line 210:
```suggestion
## WandbLogger
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+
+Finetuner allows you to utilize Weights & Biases for experiment tracking and visualization.
+The `WandBLogger` uses Weights & Biases [Anonymous Mode](https://docs.wandb.ai/ref/app/features/anon) to track a Finetuner Run.
```
scott-martens, line 213:
```suggestion
Finetuner allows you to utilize Weights & Biases for experiment tracking and visualization.
The `WandbLogger` uses Weights & Biases [Anonymous Mode](https://docs.wandb.ai/ref/app/features/anon) to track a Finetuner Run.
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+
+```{admonition} Use WandBLogger together with EvaluationCallback
+:class: hint
+The WandBLogger will track the training loss, plus the evaluation loss if `eval_data` is not None.
+
+If you use EvaluationCallback together with WandBLogger, search metrics will be tracked as well.
```
scott-martens, line 219:
```suggestion
```{admonition} Use WandbLogger together with EvaluationCallback
:class: hint
The WandbLogger will track the training loss, plus the evaluation loss if `eval_data` is not `None`.

If you use EvaluationCallback together with WandbLogger, search metrics will be tracked as well.
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+Such as `mrr`, `precision`, `recall`, etc.
+```
+
+```python
+import finetuner
+from finetuner.callback import WandBLogger, EvaluationCallback
```
scott-martens, line 225:
```suggestion
from finetuner.callback import WandbLogger, EvaluationCallback
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+
+run = finetuner.fit(
+    ...,
+    callbacks=[
+        WandBLogger(),
```
scott-martens, line 230:
```suggestion
        WandbLogger(),
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+wandb:  $ pip install wandb --upgrade
+wandb: Tracking run with wandb version 0.12.19
+wandb: Run data is saved locally in [YOUR-PATH]
+wandb: Run `wandb offline` to turn off syncing.
+wandb: Syncing run cool-wildflower-2
+wandb:  View project at https://wandb.ai/anony-mouse-279369/[YOUR-PROJECT-URL]
+wandb:  View run at https://wandb.ai/anony-mouse-279369/[YOUR-RUN-URL]
+```
+
+WandBLogger             |
```
scott-martens, line 257:
```suggestion
WandbLogger             |
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+                    __main__.py:194
+           DEBUG    Finetuning took 0 days, 0 hours 1 minutes and 16 seconds         
+                    __main__.py:196
+           INFO     Building the artifact ...                                        
+                    __main__.py:207
+           INFO     Pushing artifact to Hubble ...                                   
+                    __main__.py:231
+```
+
+## EarlyStopping
```
scott-martens, line 121:
```suggestion
## EarlyStopping

The `EarlyStopping` callback runs at the end of each training epoch and decides, based on user-provided criteria, if it should stop training and save the current model.

```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+
+Similarly to the best model checkpoint callback,
+the early stopping callback measures a given metric at the end of every epoch.
+Unlike the best model checkpoint callback,
+the early stopping callback does not save the best model;
+only the monitored metric is recorded between runs in order to assess the rate of improvement.
```
scott-martens, line 127:
```suggestion
It takes `monitor` and `mode` parameters that work the same way they do for `BestModelCheckpoint` above.
However, instead of using this information to identify the best performing model out of all epochs, it uses user-specified values at the end of each epoch to determine if it should terminate fine-tuning and save the current model.
```

guenthermi:
Maybe we could move large parts of the evaluation callback description to the evaluation section and only mention it with the parameters very shortly and refer for more details to the evaluation section. What do you think? @bwanglzu  @scott-martens

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+
```
scott-martens, line 128:
```suggestion

The user-specified parameters are:

- `min_delta`: The minimum amount of improvement that a model must have over the previous best model in order for fine-tuning to continue. By default, this is zero, meaning that the training will not stop early unless the performance starts to decrease.
- `patience`: The number of consecutive rounds without improvement before the training is stopped, set to two by default.
- `baseline`: an optional parameter that is used to compare the model's score against instead of the best previous model when checking for improvement. If specified, the improvement every epoch is measured by comparison to this value instead to the best-performing model so far.

```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+        )
+    ]
+)
+```
+
+The early stopping callback triggers at the end of training and evaluation batches to record the loss, and at the end of each epoch to evaluate the model and compare it to the best so far. Whether it stops training at the end of an epoch depends on several parameters:
+
+- `minimum_delta`: The minimum amount of improvement that a model can have over the previous best model to be considered worthwhile, zero by default, meaning that the training will not stop early unless the performance starts to decrease
+- `patience`: The number of consecutive rounds without improvement before the training is stopped, two by default.
+- `baseline`: an optional parameter that is used to compare the model's score against instead of the best previous model when checking for improvement. This baseline does not get changed over the course of a run.
```
scott-martens, line 175:
```suggestion
```
Delete because we moved info up.

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+        EarlyStopping(
+            monitor = "train_loss",
+            mode = "min",
+            patience=2,
+            min_delta=1,
+            baseline=1.5
+        )
+    ]
+)
+```
```
scott-martens, line 169:
```suggestion
```

Moved up.

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,259 @@
...
+the early stopping callback does not save the best model;
+only the monitored metric is recorded between runs in order to assess the rate of improvement.
+
+Below is some example output for a run with the early stopping callback followed by the output for the same run without the early stopping callback,
+and then the python code used to create the run.
+The output for the run with early stopping finished after just ten epochs whereas the other run finished all twenty epochs,
+resulting in nearly twice the runtime.
+That said, the resulting loss value of the early stopping run is only 0.284,
+compared to the full run's 0.272, less than five percent higher.
+The early stopping callback can be used in this way to reduce the amount of training time while still showing improvement.
```
scott-martens, line 135:
```suggestion
Below is the output of a Finetuner run with early stopping, compared to one with the same data parameters but without early stopping.
Note that the early stopping run terminated after ten epochs, while the other run continued for the whole twenty epochs. The loss of the final model is, in the two cases, nearly identical.
The early stopping run has a loss of 0.284, while taking the whole 20 epochs produced a model with a loss of 0.272.
Early stopping halved the runtime at a cost to performance of less than five percent.
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,249 @@
...
+- `baseline`: an optional parameter that is used to compare the model's score against instead of the best previous model when checking for improvement. This baseline does not get changed over the course of a run.
+
+## WiSEFTCallback
+
+WiSE-FT, proposed by Mitchell et al. in [Robust fine-tuning of zero-shot models](https://arxiv.org/abs/2109.01903),
+has been proven to be an effective way for fine-tuning models with a strong zero-shot capability,
+such as CLIP:
+
+> Large pre-trained models such as CLIP or ALIGN offer consistent accuracy across a range of data distributions when performing zero-shot inference (i.e., without fine-tuning on a specific dataset). Although existing fine-tuning methods substantially improve accuracy on a given target distribution, they often reduce robustness to distribution shifts. We address this tension by introducing a simple and effective method for improving robustness while fine-tuning: ensembling the weights of the zero-shot and fine-tuned models (WiSE-FT).
+
```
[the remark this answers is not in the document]
scott-martens: This is a quote, right?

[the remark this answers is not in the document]
bwanglzu: @scott-martens yes

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,223 @@
 (callbacks)=
-# {octicon}`link` Callbacks
\ No newline at end of file
+# {octicon}`link` Callbacks
+
+A callback is a function provided as an argument to another function that can optionally run when a specific kind of event occurs.
+There are several events during a Finetuner run that support callbacks.
```
scott-martens, line 5:
```suggestion
There are several events during a Finetuner run that support callbacks. 

```

Adding some extra spaces now that I see the built preview.

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,223 @@
...
+A callback is a function provided as an argument to another function that can optionally run when a specific kind of event occurs.
+There are several events during a Finetuner run that support callbacks. 
+You can assign callbacks to the `finetuner.fit` method with the optional `callbacks` parameter.
+
+## EvaluationCallback
+
+The `EvaluationCallback` calculates retrieval metrics at the end of each epoch for the model being tuned.
+In order to evaluate the model, two additional data sets - a query dataset and an index dataset - need to be provided as arguments.
+If no index set is provided, the evaluation is performed with the query dataset.
+To use `EvaluationCallback`:
```
scott-martens, line 13:
```suggestion

To use `EvaluationCallback`:
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,223 @@
...
+INFO     Metric: 'resnet_base_reciprocal_rank' before fine-tuning:  0.64909 after fine-tuning: 0.89224    __main__.py:210
+INFO     Metric: 'resnet_base_dcg_at_k' before fine-tuning:  1.30710 after fine-tuning: 4.52143           __main__.py:210
+INFO     Building the artifact ...                                                                        __main__.py:215
+INFO     Pushing artifact to Jina AI Cloud ...                                                            __main__.py:241
+INFO     Artifact pushed under ID ''                                                                      __main__.py:243
+DEBUG    Artifact size is 83.580 MB                                                                       __main__.py:245
+INFO     Finished 🚀    
+```
+
+For a more detailed explanation of parameters and usage, please refer to the {ref}`evaluation <evaluation>` page.
```
scott-martens, line 52:
```suggestion
For a more detailed explanation of the parameters and usage of `EvaluationCallback`, please refer to the {ref}`evaluation <evaluation>` page.
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,223 @@
...
+
+
+## BestModelCheckpoint
+
+This callback evaluates the performance of the model at the end of each epoch, and keeps a record of the best performing model across all epochs.
+Once fitting is finished the best performing model is saved instead of the most recent model.
+Finetuner determines which model is the best is based on two parameters:
+
+- `monitor`: This parameter is by default `val_loss`, which uses the evaluation data to compare models. Alternatively, you can set this to `train_loss`, which will compare models using the training data. You can specify any metric recorded by the evaluation callback for this parameter.
+- `mode`: Whether the monitored metric should be maximised (`max`) or minimised (`min`). By default the mode is set to `auto`, meaning that it will automatically choose the correct mode depending on the chosen metric: 'min' if the metric is loss and 'max' if the metric is one recorded by the evaluation callback.
```
scott-martens, line 63:
```suggestion
- `mode`: Whether the monitored metric should be maximized (`max`) or minimized (`min`). By default the mode is set to `auto`, meaning that it will automatically choose the correct mode depending on the chosen metric: `min` if the metric is loss and `max` if the metric is one recorded by the evaluation callback.
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,226 @@
...
+           DEBUG    Finetuning took 0 days, 0 hours 4 minutes and 54 seconds                             __main__.py:196
+           INFO     Building the artifact ...                                                            __main__.py:207
+           INFO     Pushing artifact to Hubble ...                                                       __main__.py:231
+```
+
+The user-specified parameters are:
+
+- `min_delta`: The minimum amount of improvement that a model must have over the previous best model in order for fine-tuning to continue. By default, this is zero, meaning that the training will not stop early unless the performance starts to decrease.
+- `patience`: The number of consecutive rounds without improvement before the training is stopped, set to two by default.
+- `baseline`: an optional parameter that is used to compare the model's score against instead of the best previous model when checking for improvement. If specified, the improvement every epoch is measured by comparison to this value instead of the best-performing model so far.
```
scott-martens, line 122:
```suggestion
- `baseline`: an optional parameter against which the model's score is compared when checking for improvement, instead of the best previous model. If specified, the improvement every epoch is measured by comparison to this value instead of the best-performing model so far.
```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,226 @@
...
+```
+
+## WiSEFTCallback
+
+WiSE-FT, proposed by Mitchell et al. in [Robust fine-tuning of zero-shot models](https://arxiv.org/abs/2109.01903),
+has proven to be an effective way of fine-tuning models with strong zero-shot capability,
+such as CLIP:
+
+> Large pre-trained models such as CLIP or ALIGN offer consistent accuracy across a range of data distributions when performing zero-shot inference (i.e., without fine-tuning on a specific dataset). Although existing fine-tuning methods substantially improve accuracy on a given target distribution, they often reduce robustness to distribution shifts. We address this tension by introducing a simple and effective method for improving robustness while fine-tuning: ensembling the weights of the zero-shot and fine-tuned models (WiSE-FT).
+
```
scott-martens, line 151:
```suggestion

To use WiSE-FT in a Finetuner run:

```

docs/concepts/callbacks.md
```diff
@@ -1,2 +1,226 @@
...
+1. When alpha is set to 0, the fine-tuned model is the same as the pre-trained model.
+2. When alpha is set to 1, the pre-trained weights are not used.
+3. When alpha is a floating-point number between 0 and 1, we combine the weights of the pre-trained and fine-tuned models.
+
+```{warning}
+It is recommended to use WiSEFTCallback when fine-tuning CLIP.
+We can not ensure it works for other types of models, such as ResNet or BERT.
+
+Please refer to {ref}`Apply WiSE-FT <wise-ft>` in the CLIP fine-tuning example for more information.
```
scott-martens, line 176: This ref seems not to work. It looks like the URL it should resolve to is: https://finetuner.jina.ai/notebooks/text_to_image/#advanced-wise-ft

LMMilliken (approved):
LGTM!

scott-martens (approved):
LGTM

PR #751, opened 2023-06-07
feat: support loading models from hf

guenthermi:
feat: support loading models from hf
---

This PR allows one to load models from https://huggingface.co/jinaai:

```python

import finetuner
import numpy as np

model = finetuner.get_model('jinaai/ecommerce-sbert-model')
e1, e2 = finetuner.encode(model, ['XBox', 'Xbox One Console 500GB - Black (2015)'])

print(f'Similarity: {np.dot(e1,e2)/(np.linalg.norm(e1)*np.linalg.norm(e2))}')
```
```bash
Similarity: 0.7842968702316284
``` 


tests/integration/test_download_hf_model.py
```diff
@@ -0,0 +1,17 @@
...
+@pytest.mark.parametrize(
+    'model',
+    [
+        'https://huggingface.co/jinaai/ecommerce-sbert-model',
+        'jinaai/ecommerce-sbert-model',
+    ],
+)
+def test_download_hf_model(model):
+    path = download_huggingface_model(model)
+    assert os.path.isdir(path)
```
LMMilliken, line 17: could we also check that we can successfully build the model from the contents of this path?

LMMilliken (changes requested):

finetuner/constants.py
```diff
@@ -94,3 +94,5 @@
 NUM_RELATIONS = 'num_relations'
 TRAIN_DATA = 'train_data'
 MAX_NUM_DOCS = 'max_num_docs'
+HF_URL_PREFIX = 'https://huggingface.co/jinaai/'
+HF_ORG_PREFIX = 'jinaai/'
```
bwanglzu, line 98: when we have the push function, let's remove the org restriction

bwanglzu (approved):
LGTM!

finetuner/__init__.py
```diff
@@ -529,12 +554,13 @@ def get_model(
...
-        specified using this argument.
-    :param token: A Jina authentication token required for pulling artifacts from
-        Hubble. If not provided, the Hubble client will try to find one either in a
-        local cache folder or in the environment.
+        directory, a path to a local zip file, a Hubble artifact ID, or a huggingface
+        model created with finetuner. Individual model artifacts (model sub-folders
+        inside the run artifacts) can also be specified using this argument.
+    :param token: A Jina authentication token (required for pulling artifacts from
+        Hubble) or a HuggingFace authentication token (required only when downloading
+        private models from huggingface) If not provided, the Hubble client might try
```
gmastrapas, line 562:
```suggestion
        private models from huggingface). If not provided, the Hubble client might try
```

finetuner/__init__.py
```diff
@@ -517,6 +519,29 @@ def build_model(
         )
 
 
+def download_huggingface_model(model_name: str, token: Optional[str] = None) -> str:
```
gmastrapas, line 522: is this meant to be public? if not, use an `_` in front of the name

finetuner/__init__.py
```diff
@@ -517,6 +519,29 @@ def build_model(
...
+    """Download a model from the HuggingFace Hub.
+
+    :param model_name: Either a URL or a model identifier from the HuggingFace Hub.
+    :param token: Optional token to access private models.
+    :return: The local path to the downloaded model.
+    """
+    from huggingface_hub import hf_hub_download, list_repo_files
```
gmastrapas, line 529: where is this requirement?

finetuner/__init__.py
guenthermi: I was not sure, maybe private is better

finetuner/__init__.py
guenthermi: we have transformers already as a dependency which depends on it. Should we add it anyway?

LMMilliken (approved):
LGTM!

